PR #876, opened 2021-08-13
Enable AKS autoscaler rule

### LINKED ISSUE #218: Enable AKS cluster autoscaler
# Rule request

## Suggested rule change

AKS clusters deployed with scale sets should enable the cluster autoscaler.

## Applies to the following

- Resource type: **Microsoft.ContainerService/managedClusters**

## Additional context

[Azure template reference](https://docs.microsoft.com/en-us/azure/templates/microsoft.containerservice/2019-08-01/managedclusters#managedclusteragentpoolprofile-object)

ArmaanMcleod:
## PR Summary

Fix #218

Added a `Azure.AKS.AutoScaling` rule for release `GA` and baseline `2021_09`. 

The rule checks boths the following types:

- `Microsoft.ContainerService/managedClusters`
- `Microsoft.ContainerService/managedClusters/agentPools`

Also only checks agent pools of type `VirtualMachineScaleSets`, and agent pools that are any other type(such as `AvailabilitySets` just pass. Correct me if this needs to be handled differently. 

Also let me know if I should provide more detail in `docs/en/rules/Azure.AKS.AutoScaling.md`.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.AKS.AutoScaling.md
+++ b/docs/en/rules/Azure.AKS.AutoScaling.md
@@ -1,0 +1,141 @@
+---
+severity: Important
+pillar: Performance Efficiency
+category: Performance
+resource: Azure Kubernetes Service
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AKS.AutoScaling/
+---
+
+# Enable AKS cluster autoscaler
+
+## SYNOPSIS
+
+Use Autoscaling to ensure AKS clusters deployed with virtual machine scale sets are running efficiently with the right number of nodes for the workloads present.
+
+## DESCRIPTION
+
+In addition to perform manual scaling, AKS clusters support autoscaling.
+Autoscaling reduces manual intervention required to scale a cluster to keep up with application demands.
+
+## RECOMMENDATION
+
+Consider enabling autoscaling for AKS clusters deployed with virtual machine scale sets.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To set enable autoscaling for an AKS cluster:
+
+- Set `properties.agentPoolProfiles[*].enableAutoScaling` to `true`.
+- Set `properties.agentPoolProfiles[*].type` to `VirtualMachineScaleSets`.
+
+For example:
+
+```json
+{
+    "comments": "Azure Kubernetes Cluster",
+    "apiVersion": "2020-12-01",
+    "dependsOn": [
+        "[resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', parameters('identityName'))]"
+    ],
+    "type": "Microsoft.ContainerService/managedClusters",
+    "location": "[parameters('location')]",
+    "name": "[parameters('clusterName')]",
+    "identity": {
+        "type": "UserAssigned",
+        "userAssignedIdentities": {
+            "[resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', parameters('identityName'))]": {}
+        }
+    },
+    "properties": {
+        "kubernetesVersion": "[parameters('kubernetesVersion')]",
+        "disableLocalAccounts": true,
+        "enableRBAC": true,
+        "dnsPrefix": "[parameters('dnsPrefix')]",
+        "agentPoolProfiles": [
+            {
+                "name": "system",
+                "osDiskSizeGB": 32,
+                "count": 3,
+                "minCount": 3,
+                "maxCount": 10,
+                "enableAutoScaling": true,
+                "maxPods": 50,
+                "vmSize": "Standard_D2s_v3",
+                "osType": "Linux",
+                "type": "VirtualMachineScaleSets",
+                "vnetSubnetID": "[variables('clusterSubnetId')]",
+                "mode": "System",
+                "osDiskType": "Ephemeral",
+                "scaleSetPriority": "Regular"
+            }
+        ],
+        "aadProfile": {
+            "managed": true,
+            "enableAzureRBAC": true,
+            "adminGroupObjectIDs": "[parameters('clusterAdmins')]",
+            "tenantID": "[subscription().tenantId]"
+        },
+        "networkProfile": {
+            "networkPlugin": "azure",
+            "networkPolicy": "azure",
+            "loadBalancerSku": "Standard",
+            "serviceCidr": "192.168.0.0/16",
+            "dnsServiceIP": "192.168.0.4",
+            "dockerBridgeCidr": "172.17.0.1/16"
+        },
+        "autoUpgradeProfile": {
+            "upgradeChannel": "stable"
+        },
+        "addonProfiles": {
+            "azurepolicy": {
+                "enabled": true,
+                "config": {
+                    "version": "v2"
+                }
+            },
+            "omsagent": {
+                "enabled": true,
+                "config": {
+                    "logAnalyticsWorkspaceResourceID": "[parameters('workspaceId')]"
+                }
+            },
+            "kubeDashboard": {
+                "enabled": false
+            }
+        }
+    }
+}
+```
+
+### Configure with Azure CLI
+
+#### Enable cluster autoscaler
+
+```bash
+az aks update \
+  --name '<name>' \
+  --resource-group '<resource_group>' \
+  --enable-cluster-autoscaler \
+  --min-count '<min_count>' \
+  --max-count '<max_count>'
+```
+
+#### Enable cluster nodepool autoscaler
+
+```bash
+az aks nodepool update \
+  --name '<name>' \
+  --resource-group '<resource_group>' \
+  --cluster-name '<cluster_name>' \
+  --enable-cluster-autoscaler \
+  --min-count '<min_count>' \
+  --max-count '<max_count>'
+```
+
+## LINKS
+
+- [Automatically scale a cluster to meet application demands on Azure Kubernetes Service (AKS)](https://docs.microsoft.com/en-us/azure/aks/cluster-autoscaler)
+- [Scaling options for applications in Azure Kubernetes Service (AKS)](https://docs.microsoft.com/en-us/azure/aks/concepts-scale)
+- [Azure template reference](https://docs.microsoft.com/azure/templates/microsoft.containerservice/managedclusters)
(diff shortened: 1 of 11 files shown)


docs/en/rules/Azure.AKS.AutoScaling.md
```diff
@@ -0,0 +1,141 @@
+---
+severity: Important
+pillar: Performance Efficiency
+category: Performance
+resource: Azure Kubernetes Service
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.AKS.AutoScaling/
```
ArmaanMcleod, line 6: Ah just realized this will be 404

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -138,6 +138,26 @@ Rule 'Azure.AKS.AzureRBAC' -Type 'Microsoft.ContainerService/managedClusters' -T
...
+    $agentPools = @(GetAgentPoolProfiles);
+
+    if ($agentPools.Length -eq 0) {
+        return $Assert.Pass();
+    }
+
+    foreach ($agentPool in $agentPools) {
+
+        # Autoscaling only available on virtual machine scale sets
+        if ($Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets')) {
```
BernieWhite, line 152: I think this line is not going to be performing how you intended.

`$Assert.HasFieldValue` is going to be returning an `AssertResult` which is always an object, and will also pass the null check on this line.
To get the result of `HasFieldValue`, use `$Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets').Result` or explictilty cast to a boolean. i.e. `[bool]$Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets')`. Personally I prefer the `.Result` option because it is clearer.

This is why `cluster-A`, and `cluster-B` in the unit tests are reporting as fail instead of pass.

The approach of only reporting on scale set node pools make sense to me.

tests/PSRule.Rules.Azure.Tests/Azure.AKS.Tests.ps1
```diff
@@ -230,6 +230,22 @@ Describe 'Azure.AKS' -Tag AKS {
...
         }
 
+        It 'Azure.AKS.AutoScaling' {
+            $filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.AKS.AutoScaling' };
+
+            # Fail
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult | Should -HaveCount 4
+            $ruleResult.TargetName | Should -BeIn 'cluster-A', 'cluster-B', 'cluster-C', 'cluster-D';
```
BernieWhite, line 240: In think cluster A, B should be returning Pass. Altnatively you could filter out clusters/ node pools that have no scale set node pools by using a `-If` pre-condition.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -162,6 +182,7 @@ function global:GetAgentPoolProfiles {
                 type = $TargetObject.properties.type
                 maxPods = $TargetObject.properties.maxPods
                 orchestratorVersion = $TargetObject.properties.orchestratorVersion
+                enableAutoScaling = $TargetObject.properties.enableAutoScaling
```
BernieWhite, line 185: I think this will also need to be inserted after line 175 `orchestratorVersion = $_.properties.orchestratorVersion` above.

Doesn't look like there is test coverage for this but the scenario where the agentPool is a nested sub-resource in the cluster i.e. In the `"resources": [ ]` of the cluster should ideally work as well.

BernieWhite (changes requested):
@ArmaanMcleod Great work! A few suggestions but otherwise all good.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: @BernieWhite Ah that makes sense, I had a look at the `AssertRule` class and didn't see that Result property: https://github.com/microsoft/PSRule/blob/main/src/PSRule/Runtime/AssertResult.cs#L33-L36

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: This was another thing I was unsure about. If you have sub resources that are invalid, the rule record still outcome shows up at `Pass`.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: Should be fixed now

tests/PSRule.Rules.Azure.Tests/Azure.AKS.Tests.ps1
ArmaanMcleod: Fixed

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: Yeah so tested this with this template: 

```json
    "resources": [
        {
            "apiVersion": "2021-02-01",
            "dependsOn": [
                "[concat('Microsoft.Resources/deployments/', 'WorkspaceDeployment-20210814190241')]"
            ],
            "type": "Microsoft.ContainerService/managedClusters",
            "location": "[parameters('location')]",
            "name": "[parameters('resourceName')]",
            "properties": {
                "kubernetesVersion": "[parameters('kubernetesVersion')]",
                "enableRBAC": "[parameters('enableRBAC')]",
                "dnsPrefix": "[parameters('dnsPrefix')]",
                "agentPoolProfiles": [
                    {
                        "name": "agentpool1",
                        "osDiskSizeGB": "[parameters('osDiskSizeGB')]",
                        "count": 3,
                        "enableAutoScaling": true,
                        "minCount": 1,
                        "maxCount": 5,
                        "vmSize": "Standard_DS2_v2",
                        "osType": "Linux",
                        "storageProfile": "ManagedDisks",
                        "type": "VirtualMachineScaleSets",
                        "mode": "System",
                        "maxPods": 110,
                        "availabilityZones": null
                    }
                ],
                "networkProfile": {
                    "loadBalancerSku": "standard",
                    "networkPlugin": "[parameters('networkPlugin')]"
                },
                "apiServerAccessProfile": {
                    "enablePrivateCluster": "[parameters('enablePrivateCluster')]"
                },
                "addonProfiles": {
                    "httpApplicationRouting": {
                        "enabled": "[parameters('enableHttpApplicationRouting')]"
                    },
                    "azurepolicy": {
                        "enabled": "[parameters('enableAzurePolicy')]"
                    },
                    "omsAgent": {
                        "enabled": false
                    }
                }
            },
            "tags": {},
            "identity": {
                "type": "SystemAssigned"
            },
            "resources": [
                {
                    "type": "Microsoft.ContainerService/managedClusters/agentPools",
                    "apiVersion": "2021-02-01",
                    "name": "[concat(parameters('resourceName'), '/agentpool2')]",
                    "dependsOn": [
                        "[resourceId('Microsoft.ContainerService/managedClusters', parameters('resourceName'))]"
                    ],
                    "properties": {
                        "osDiskSizeGB": "[parameters('osDiskSizeGB')]",
                        "count": 3,
                        "enableAutoScaling": false,
                        "vmSize": "Standard_DS2_v2",
                        "osType": "Linux",
                        "storageProfile": "ManagedDisks",
                        "type": "VirtualMachineScaleSets",
                        "mode": "System",
                        "maxPods": 110,
                        "availabilityZones": null
                    }
                }
            ]
        }
    ]
```

Which gave the following:

```pwsh

    ____  _____ ____        __
   / __ \/ ___// __ \__  __/ /__
  / /_/ /\__ \/ /_/ / / / / / _ \
 / ____/___/ / _, _/ /_/ / /  __/
/_/    /____/_/ |_|\__,_/_/\___/

Using PSRule v1.6.0+473073bb8bfb8cc11051923545e290f916850110
Using PSRule.Rules.Azure v0.0.1

----------------------------
Explore documentation: https://microsoft.github.io/PSRule/
Contribute and find source: https://github.com/microsoft/PSRule
Report issues: https://github.com/microsoft/PSRule/issues
PSRule.Rules.Azure: https://aka.ms/ps-rule-azure
----------------------------

> akstestcluster : Microsoft.ContainerService/managedClusters [1/1]

   PASS  Azure.AKS.AutoScaling

Rules processed: 1, failed: 0, errored: 0
```

Seems like I need to check how these sub resources are being checked since this doesn't look right.

ArmaanMcleod: I guess another challenge here is that I'm unable to debug this by adding breakpoints inside the rules. I usually add this VSCODE configuration to debug PowerShell interactively:

```json
{
    "name": "PowerShell: Interactive Session",
    "type": "PowerShell",
    "request": "launch",
    "cwd": ""
}
```

But doesn't seem to have any effect if I place breakpoints inside rules.

ArmaanMcleod: I guess another issue I'm having is that I can't set breakpoints within the rules and step through with the VSCode debugger. I ussually debug these issues with Interactive sessions but for some reason when I run `Assert-PSRule` it doesn't pick them up.

BernieWhite: Your example template looks correct to me.

I think the errors is related to line 173, `type = $_.type` which should be `type = $_.properties.type`.
This appears to be a bug with the existing code.

Because the node pools is not matching the scale set type it's being passed.

BernieWhite (commented):
@ArmaanMcleod Nice. Almost there. One minor tweak and all good.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: Ah good catch, that went past me. Should be fixed now.

ArmaanMcleod:
@BernieWhite Thanks for the help 🙂 . I fixed the typo in `GetAgentProfiles` and also added another cluster to test sub resources. Seems to be working fine now.

BernieWhite (approved):
@ArmaanMcleod All good.

PR #906, opened 2021-08-23
AKS Azure AvailabilityZone Rule

### LINKED ISSUE #880: AKS clusters should use availability zones for regions that support it
# Rule request

## Suggested rule change

AKS clusters support availability zones. Clusters deployed into regions that also support availability zones should have this configured.

## Applies to the following

- Resource type: **Microsoft.ContainerService/managedClusters**

ArmaanMcleod:
## PR Summary

Fix #880

Added a `Azure.AKS.AvailabilityZone` option for the following types:

- Microsoft.ContainerService/managedClusters

Added the zone mappings from https://docs.microsoft.com/en-us/rest/api/resources/providers/list to `providers.json` with the regions that support availability zones. Rule configures extra zones with the `AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST` configuration option

Also did a bit of refactoring and made a `ResourceProviderHelper` class to share those common methods to read the `providers.json` file. Let me know if you'd like this to be different. I didn't want to copy the methods over to the `Helper` class since it seems like these methods could be reused a lot if interacting with `provders.json` data. 

Also added some tests but can add more to avoid regression in the future. There is a couple of edge cases where the zones passed in from the custom configuration could be incorrect, but didn't know how this should be handled. I can probably add a check to make sure valid zones are being passed in.

Let me know what you think of this so far 🙂 

## PR Checklist

- [x] Summarized changes
- [ ] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -60,5 +60,8 @@
         "text",
         "yaml",
         "yml"
-    ]
+    ],
+    "[csharp]": { 
+        "editor.formatOnSave": true
+    }
 }
--- a/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
@@ -177,7 +177,7 @@ Rule 'Azure.AKS.CNISubnetSize' -Type 'Microsoft.ContainerService/managedClusters
 Rule 'Azure.AKS.AvailabilityZone' -Type 'Microsoft.ContainerService/managedClusters', 'Microsoft.ContainerService/managedClusters/agentPools' -Tag @{ release = 'GA'; ruleSet = '2021_09'; } {
     $agentPools = @(GetAgentPoolProfiles);
 
-    if ($agentPools.Length -eq 0) {
+    if ($agentPools.Length -eq 0 -or [string]::IsNullOrEmpty($TargetObject.Location)) {
         return $Assert.Pass();
     }
 
@@ -198,7 +198,7 @@ Rule 'Azure.AKS.AvailabilityZone' -Type 'Microsoft.ContainerService/managedClust
     | Where-Object { (GetNormalLocation -Location $_.Location) -eq $location } 
     | Select-Object -ExpandProperty Zones -First 1;
 
-    if (!$locationAvailabilityZones) {
+    if (-not $locationAvailabilityZones) {
         return $Assert.Pass();
     }
 
--- a/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
@@ -173,6 +173,50 @@ Rule 'Azure.AKS.CNISubnetSize' -Type 'Microsoft.ContainerService/managedClusters
     }
 } -Configure @{ AZURE_AKS_CNI_MINIMUM_CLUSTER_SUBNET_SIZE = 23 }
 
+# Synopsis: AKS clusters deployed with virtual machine scale sets should use availability zones in supported regions for high availability.
+Rule 'Azure.AKS.AvailabilityZone' -Type 'Microsoft.ContainerService/managedClusters', 'Microsoft.ContainerService/managedClusters/agentPools' -Tag @{ release = 'GA'; ruleSet = '2021_09'; } {
+    $agentPools = @(GetAgentPoolProfiles);
+
+    if ($agentPools.Length -eq 0) {
+        return $Assert.Pass();
+    }
+
+    $configurationZones = $Configuration.AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST;
+
+    $virtualMachineScaleSetProvider = [PSRule.Rules.Azure.Runtime.Helper]::GetResourceType('Microsoft.Compute', 'virtualMachineScaleSets');
+
+    if ($configurationZones.Length -gt 0) {
+        $availabilityZones = @($configurationZones) + $virtualMachineScaleSetProvider.ZoneMappings;
+    }
+    else {
+        $availabilityZones = $virtualMachineScaleSetProvider.ZoneMappings;
+    }
+
+    $location = GetNormalLocation -Location $TargetObject.Location;
+
+    $locationAvailabilityZones = $availabilityZones 
+    | Where-Object { (GetNormalLocation -Location $_.Location) -eq $location } 
+    | Select-Object -ExpandProperty Zones -First 1;
+
+    if (!$locationAvailabilityZones) {
+        return $Assert.Pass();
+    }
+
+    $joinedZoneString = $locationAvailabilityZones -join ', ';
+
+    foreach ($agentPool in $agentPools) {
+
+        # Availability zones only available on virtual machine scale sets
+        if ($Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets').Result) {
+            $validZone = $Assert.HasField($agentPool, 'availabilityZones').Result -and -not $Assert.NullOrEmpty($agentPool, 'availabilityZones').Result;
+            $Assert.Create($validZone, ($LocalizedData.AKSAvailabilityZone -f $agentPool.name, $location, $joinedZoneString));
+        }
+        else {
+            $Assert.Pass();
+        }
+    }
+} -Configure @{ AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST = @() }
+
 #region Helper functions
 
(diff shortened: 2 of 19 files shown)


.vscode/settings.json
```diff
@@ -60,5 +60,8 @@
         "text",
         "yaml",
         "yml"
-    ]
+    ],
+    "[csharp]": {
```
ArmaanMcleod, line 64: Let me know if this ok to add in. I ussually have this always set in VSCode when making changes to csharp files.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -173,6 +173,50 @@ Rule 'Azure.AKS.CNISubnetSize' -Type 'Microsoft.ContainerService/managedClusters
...
+        # Availability zones only available on virtual machine scale sets
+        if ($Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets').Result) {
+            $validZone = $Assert.HasField($agentPool, 'availabilityZones').Result -and -not $Assert.NullOrEmpty($agentPool, 'availabilityZones').Result;
+            $Assert.Create($validZone, ($LocalizedData.AKSAvailabilityZone -f $agentPool.name, $location, $joinedZoneString));
+        }
+        else {
+            $Assert.Pass();
+        }
+    }
+} -Configure @{ AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST = @() }
```
ArmaanMcleod, line 218: I probably need to test this with some custom zones. Need to check if my format in `/docs/setup/configuring-rules.md` will get serialized properly into an array of `PSCustomObject` with yamldotnet.

BernieWhite: Yeah, we should ideally have a test case for custom configuration. Objects should be serialized as PSObject.

BernieWhite (changes requested):
@ArmaanMcleod Nice work! Some suggested changes but otherwise all good.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: Thanks @BernieWhite. Is there a way to easily test this? Or do I have to create a `ps-rule.yml` file with these custom zones. Would be cool to be able to pass in these configurations via a `-Configuration` switch that accepts a hashtable or something.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -173,6 +173,48 @@ Rule 'Azure.AKS.CNISubnetSize' -Type 'Microsoft.ContainerService/managedClusters
     }
 } -Configure @{ AZURE_AKS_CNI_MINIMUM_CLUSTER_SUBNET_SIZE = 23 }
 
+# Synopsis: AKS clusters deployed with virtual machine scale sets should use availability zones in supported regions for high availability.
+Rule 'Azure.AKS.AvailabilityZone' -Type 'Microsoft.ContainerService/managedClusters', 'Microsoft.ContainerService/managedClusters/agentPools' -Tag @{ release = 'GA'; ruleSet = '2021_09'; } {
```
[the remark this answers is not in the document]
ArmaanMcleod: Fixed

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -173,6 +173,48 @@ Rule 'Azure.AKS.CNISubnetSize' -Type 'Microsoft.ContainerService/managedClusters
...
+        return $Assert.Pass();
+    }
+
+    $joinedZoneString = $locationAvailabilityZones -join ', ';
+
+    foreach ($agentPool in $agentPools) {
+
+        # Availability zones only available on virtual machine scale sets
+        if ($Assert.HasFieldValue($agentPool, 'type', 'VirtualMachineScaleSets').Result) {
+            $validZone = $Assert.HasField($agentPool, 'availabilityZones').Result -and -not $Assert.NullOrEmpty($agentPool, 'availabilityZones').Result;
```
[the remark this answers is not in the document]
ArmaanMcleod: Fixed

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
ArmaanMcleod: I know in pester I can use `TestDrive` to create files just for testing, perhaps I could use that here and dump some YAML config in there.

BernieWhite: @ArmaanMcleod There is a few ways to provide options:

- Pass a hashtable. See [example](https://github.com/Azure/PSRule.Rules.Azure/blob/6ecd4e7cd369f4261780a823df6826241681acc7/tests/PSRule.Rules.Azure.Tests/Azure.Convention.Tests.ps1#L49-L52)
- Pass a path to a options file. [see](https://microsoft.github.io/PSRule/concepts/PSRule/en-US/about_PSRule_Options.html)
- Pass an options object. [example](https://github.com/Azure/PSRule.Rules.Azure/blob/6ecd4e7cd369f4261780a823df6826241681acc7/tests/PSRule.Rules.Azure.Tests/Azure.Resource.Tests.ps1#L92-L100)

ArmaanMcleod: Thanks @BernieWhite. Hashtable works fine. However when I use this config:

```yaml
# PSRule options for unit testing

configuration:
  AZURE_SUBSCRIPTION:
    subscriptionId: 'ffffffff-ffff-ffff-ffff-ffffffffffff'
    tenantId: 'ffffffff-ffff-ffff-ffff-ffffffffffff'
    displayName: 'Unit Test Subscription'
    state: 'NotDefined'
  AZURE_RESOURCE_GROUP:
    name: 'ps-rule-test-rg'
    location: 'eastus'
    tags:
      env: 'prod'
    properties:
      provisioningState: 'Succeeded'
  AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST:
  - location: 'Australia Southeast'
    zones:
      - "1"
      - "2"
      - "3"
  - location: 'Norway East'
    zones:
      - "1"
      - "2"
      - "3"
```

Doesn't seem to work and can't read the `$_.Location` property on line 197. 

I am just testing this with:

```pwsh
$configPath = Join-Path -Path $here -ChildPath 'ps-rule-options.yaml';
$result = Invoke-PSRule @invokeParams -InputPath $dataPath -Option $configPath
```

Seems to me its not liking the YAML format.

BernieWhite: @ArmaanMcleod Seems to be a bug. The object is being deserialized as a dictionary.

```powershell
$o = New-PSRuleOption .\tests\PSRule.Rules.Azure.Tests\ps-rule-options.yaml
$o.Configuration['AZURE_AKS_ADDITIONAL_REGION_AVAILABILITY_ZONE_LIST']
```

returns:

```text
Key      Value
---      -----
location Australia Southeast
zones    {1, 2, 3}
location Norway East
zones    {1, 2, 3}
```

ArmaanMcleod: @BernieWhite Yeah I was able to reproduce the same just then. I guess the workaround could be to convert the dictionary to an array inside the rule, or change the YAML fomat to something else. I stuck with the array since that's how the API exports the provider zoneMappings, and just kept it consistent for the option.

BernieWhite: @ArmaanMcleod Try using pre-release v1.7.0-B2108021 in 15mins. Just pushing a release now.

ArmaanMcleod: @BernieWhite That was quick, thanks for that 😄 . I've tested the pre-release locally and it seems to be working fine with the tests I've added. Will see if it passes CI checks.

ArmaanMcleod: Yeah tests failing in CI since its using latest stable 1.6.0 PSRule module.

BernieWhite (approved):
@ArmaanMcleod All good. Nice work. I'll push out an interim patch for v1.6.0. And we can rerun the CI for merging.

PR #973, opened 2021-09-17
Added load balancer Standard SKU rule

### LINKED ISSUE #957: Load balancers should use Standard SKU
# Rule request

## Suggested rule change

Load balancers should use Standard SKU for production workloads.

## Applies to the following

- Resource type: Microsoft.Network/loadBalancers

## Additional context

Related to #927

ArmaanMcleod:
## PR Summary

Fix #957

Added `Azure.LB.StandardSKU` rule for the following types:

- Microsoft.Network/loadBalancers

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
@@ -243,6 +243,11 @@ Rule 'Azure.LB.AvailabilityZone' -Type 'Microsoft.Network/loadBalancers' -If { I
     }
 }
 
+# Synopsis: Load balancers should be deployed with Standard SKU for production workloads.
+Rule 'Azure.LB.StandardSKU' -Type 'Microsoft.Network/loadBalancers' -Tag @{ release = 'GA'; ruleSet = '2021_09'; } {
+    IsStandardLoadBalancer;
+}
+
 #endregion Load Balancer
 
 #region Azure Firewall
--- a/src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
@@ -228,7 +228,7 @@ Rule 'Azure.LB.Name' -Type 'Microsoft.Network/loadBalancers' -Tag @{ release = '
 }
 
 # Synopsis: Load balancers deployed with Standard SKU should be zone-redundant for high availability.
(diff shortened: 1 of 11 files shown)


src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
```diff
@@ -243,6 +243,11 @@ Rule 'Azure.LB.AvailabilityZone' -Type 'Microsoft.Network/loadBalancers' -If { I
     }
 }
 
+# Synopsis: Load balancers should be deployed with Standard SKU for production workloads.
+Rule 'Azure.LB.StandardSKU' -Type 'Microsoft.Network/loadBalancers' -Tag @{ release = 'GA'; ruleSet = '2021_09'; } {
+    IsStandardLoadBalancer;
```
BernieWhite, line 248: By itself this would only return true which would not provide a good reason.

Nothing wrong with calling `IsStandardLoadBalancer` but maybe update function to remove `.Result` so the reason is returned. This should still work for the other pre-condition using this function.

BernieWhite (changes requested):
@ArmaanMcleod Thanks Armaan. One suggested change. Otherwise, all good to merge.

src/PSRule.Rules.Azure/rules/Azure.VNET.Rule.ps1
ArmaanMcleod: Ah good catch @BernieWhite. Fixed 🙂

ArmaanMcleod: Ah I might change the `IsStandardLoadBalancer` output type to `PSRule.Runtime.AssertResult`.

ArmaanMcleod:
@BernieWhite should be all good to go

BernieWhite (approved):

BernieWhite (approved):
@ArmaanMcleod All good. Nice work. All good to merge.

Strictly speaking `-If { (IsStandardLoadBalancer).Result }` should not be required. Not requesting a change, just letting you know that PSRule should automatically resolve an AssertResult within a pre-condition so adding `.Result` is not required.

ArmaanMcleod:
> @ArmaanMcleod All good. Nice work. All good to merge.
> 
> Strictly speaking `-If { (IsStandardLoadBalancer).Result }` should not be required. Not requesting a change, just letting you know that PSRule should automatically resolve an AssertResult within a pre-condition so adding `.Result` is not required.

@BernieWhite All good. I've removed .Result from the precondition.

PR #983, opened 2021-09-20
Added Public IP address Availability Zone and Standard SKU rules

ArmaanMcleod:
## PR Summary

Fix #958
Fix #979 

Added the following rules:

- `Azure.PublicIP.AvailabilityZone`
- `Azure.PublicIP.StandardSKU`

for the following types:

- Microsoft.Network/publicIPAddresses

## PR Checklist

- [x] PR has a meaningful title
- [x] Summarized changes
- [x] Change is not breaking
- [x] This PR is ready to merge and is not **Work in Progress**
- **Rule changes**
  - [x] Unit tests created/ updated
  - [x] Rule documentation created/ updated
  - [x] Link to a filed issue
  - [x] [Change log](https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/CHANGELOG-v1.md) has been updated with change under unreleased section

PR #984, opened 2021-09-22
Use REST API to fetch zones instead of Get-AzPublicIpAddress cmdlet

ArmaanMcleod:
## PR Summary

Apologise for not bringing this up in the previous PR, but I pulled out some code that I stashed that uses the Rest API to get the zones instead of the `Get-AzPublicIpAddress` cmdlet. 

Thought it was a good idea to suggest this instead since we can remove the `Az.Network` dependency and have some cmdlets available to fetch resources from API when this issue happens again. This prevents the need to bring in other Az module dependencies. 

Let me know what you think. Perfectly happy to not merge this is the current approach is fine with you 🙂. 

To me it seems better to do this given alot of resources from `Get-AzResource` have missing properties, and means we don't need to introduce new Az modules everytime and just call the API instead. 

Related: https://github.com/Azure/PSRule.Rules.Azure/pull/983

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/PSRule.Rules.Azure.psd1
+++ b/src/PSRule.Rules.Azure/PSRule.Rules.Azure.psd1
@@ -53,7 +53,7 @@ DotNetFrameworkVersion = '4.7.2'
 # Modules that must be imported into the global environment prior to importing this module
 RequiredModules = @(
     @{ ModuleName = 'PSRule'; ModuleVersion = '0.0.1' }
-    @{ ModuleName = 'Az.Accounts'; ModuleVersion = '1.5.2' }
+    @{ ModuleName = 'Az.Accounts'; ModuleVersion = '2.5.2' }
     @{ ModuleName = 'Az.Resources'; ModuleVersion = '1.4.0' }
 )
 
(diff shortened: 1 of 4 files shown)


BernieWhite (commented):
@ArmaanMcleod Great improvement. Yes, ideally let's not take a dependency if possible because it will take longer to install the dependency chain.

Alternatively, though maybe we could use `Get-AzRest`, `Get-AzRestMethod`, or `Get-AzAccessToken` since these are exposed as cmdlets they are less likely to break then calling internal class methods which could be refactored.

ArmaanMcleod:
@BernieWhite Thats a great suggestion. I've converted the code to use `Invoke-AzRestMethod` which is much cleaner.

src/PSRule.Rules.Azure/PSRule.Rules.Azure.psd1
```diff
@@ -55,7 +55,6 @@ RequiredModules = @(
     @{ ModuleName = 'PSRule'; ModuleVersion = '0.0.1' }
     @{ ModuleName = 'Az.Accounts'; ModuleVersion = '1.5.2' }
```
BernieWhite, line 56: @ArmaanMcleod Lets bump this `Az.Accounts` version. `Invoke-AzRestMethod` was added in v1.9.1 to at least this version. We also use v2.5.2 in the build pipeline so that would be ok too.

BernieWhite (approved):
@ArmaanMcleod Looks good. One suggestion but otherwise good to merge.

src/PSRule.Rules.Azure/PSRule.Rules.Azure.psd1
ArmaanMcleod: @BernieWhite Good point. I've bumped the version to 2.5.2 🙂

BernieWhite (approved):
@ArmaanMcleod All good. :)

PR #1141, opened 2021-12-14
Added rules to check comments/descriptions for each template resource

### LINKED ISSUE #969: Use comments for each template resource
# Rule request

## Suggested rule change

Specify comments for each resource included within an Azure template to communicate the purpose of the resource.

## Applies to the following

- Azure templates

## Additional context

[Reference](https://docs.microsoft.com/azure/azure-resource-manager/templates/best-practices#resources)

ArmaanMcleod:
## PR Summary

Fix #969

Added `Azure.Template.UseComments` and `Azure.Template.UseDescriptions` rules.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
@@ -312,8 +312,19 @@ Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) } -Tag @{
         return $Assert.Pass();
     }
 
+    # ARM template
+    if (-not(IsGenerated)) {
+        $commentField = 'comments';
+    }
+
+    # Generated template e.g. bicep
+    # Can check description in metadata
+    else {
+        $commentField = 'metadata.description';
+    }
+
     $numResourcesWithoutComments = @($resources | Where-Object {
-        -not ($Assert.HasField($_, 'comments').Result) -or $Assert.NullOrEmpty($_, 'comments').Result
+        -not ($Assert.HasField($_, $commentField).Result) -or $Assert.NullOrEmpty($_, $commentField).Result
     }).Length
 
     $Assert.Create(
--- a/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
@@ -312,9 +312,16 @@ Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) } -Tag @{
         return $Assert.Pass();
     }
 
-    foreach ($resource in $resources) {
-        $Assert.HasField($resource, 'comments');
-    }
+    $numResourcesWithoutComments = @($resources | Where-Object {
+        -not ($Assert.HasField($_, 'comments').Result) -or $Assert.NullOrEmpty($_, 'comments').Result
+    }).Length
+
+    $Assert.Create(
+        ($numResourcesWithoutComments -eq 0),
+        $LocalizedData.TemplateResourceWithoutComment,
+        $TargetObject.FullName,
+        $numResourcesWithoutComments
+    )
 }
 
 #endregion Parameters
--- a/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
@@ -304,40 +304,63 @@ Rule 'Azure.Template.ValidSecretRef' -Type '.json' -If { (IsParameterFile) } -Ta
     }
 }
 
-# Synopsis: Use comments for each resource to communicate purpose.
-Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) } -Tag @{ release = 'GA'; ruleSet = '2021_12'; } {
+# Synopsis: Use comments for each resource in ARM template to communicate purpose.
+Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) -and !(IsGenerated) } -Tag @{ release = 'GA'; ruleSet = '2021_12'; } {
     $resources = @(GetTemplateResources);
 
     if ($resources.Length -eq 0) {
         return $Assert.Pass();
     }
 
-    # ARM template
-    if (-not(IsGenerated)) {
-        $commentField = 'comments';
-    }
-
-    # Generated template e.g. bicep
-    # Can check description in metadata
-    else {
-        $commentField = 'metadata.description';
-    }
-
-    $numResourcesWithoutComments = @($resources | Where-Object {
-        -not ($Assert.HasField($_, $commentField).Result) -or $Assert.NullOrEmpty($_, $commentField).Result
-    }).Length
+    $numResourcesWithoutComments = GetNumberOfResourcesWithoutField -Resource $resources -Field 'comments';
 
     $Assert.Create(
         ($numResourcesWithoutComments -eq 0),
         $LocalizedData.TemplateResourceWithoutComment,
         $TargetObject.FullName,
         $numResourcesWithoutComments
-    )
+    );
+}
+
+# Synopsis: Use descriptions for each resource in generated template(bicep, psarm, AzOps) to communicate purpose.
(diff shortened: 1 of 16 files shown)


tests/PSRule.Rules.Azure.Tests/Azure.Template.Tests.ps1
```diff
@@ -655,5 +655,39 @@ Describe 'Azure.Template' -Tag 'Template' {
...
+            $filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.Template.UseComments' };
+
+            # Fail
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult.Length | Should -Be 3;
+            $ruleResult.TargetName | Should -BeLike "*Resources.Template[2-4].json";
+
+            $ruleResult[0].Reason | Should -HaveCount 2;
+            $ruleResult[0].Reason[0] | Should -BeExactly "The field 'comments' does not exist.";
```
ArmaanMcleod, line 671: Was seeing if we could put resource name in the reasons, but this doesn't work out when we use ARM template expressions like `[concat(parameters('vnetName'), '/', parameters('clusterSubnet'), '/Microsoft.Authorization/', guid(parameters('clusterSubnet')))]`.

tests/PSRule.Rules.Azure.Tests/Azure.Template.Tests.ps1
```diff
@@ -655,5 +655,39 @@ Describe 'Azure.Template' -Tag 'Template' {
...
+            $ruleResult[0].Reason[1] | Should -BeExactly "The field 'comments' does not exist.";
+
+            $ruleResult[1].Reason | Should -HaveCount 4;
+            $ruleResult[1].Reason[0] | Should -BeExactly "The field 'comments' does not exist.";
+            $ruleResult[1].Reason[1] | Should -BeExactly "The field 'comments' does not exist.";
+            $ruleResult[1].Reason[2] | Should -BeExactly "The field 'comments' does not exist.";
+            $ruleResult[1].Reason[3] | Should -BeExactly "The field 'comments' does not exist.";
```
ArmaanMcleod, line 678: Was seeing if we could put resource name in the reasons, but this doesn't work out when we use ARM template expressions like `[concat(parameters('vnetName'), '/', parameters('clusterSubnet'), '/Microsoft.Authorization/', guid(parameters('clusterSubnet')))]`, since that is what gets shown in the reason if I do something like `.Reason($LocalizedData.UseComments, $resource.name)` in the rule.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
```diff
@@ -304,6 +304,19 @@ Rule 'Azure.Template.ValidSecretRef' -Type '.json' -If { (IsParameterFile) } -Ta
...
+Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) } -Tag @{ release = 'GA'; ruleSet = '2021_12'; } {
+    $resources = @(GetTemplateResources);
+
+    if ($resources.Length -eq 0) {
+        return $Assert.Pass();
+    }
+
+    foreach ($resource in $resources) {
+        $Assert.HasField($resource, 'comments');
+    }
```
ArmaanMcleod, line 317: This is where a LINQ `All()` equivalent for PSRule would be usefule.

ArmaanMcleod:
Strange how the tests are failing for macOS.

tests/PSRule.Rules.Azure.Tests/Azure.Template.Tests.ps1
BernieWhite: > Strange how the tests are failing for macOS.

This likely the order the files are being processed, which is different from Windows. Filter `$ruleResult` by target name before you compare.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
```diff
@@ -304,6 +304,19 @@ Rule 'Azure.Template.ValidSecretRef' -Type '.json' -If { (IsParameterFile) } -Ta
...
+# Synopsis: Use comments for each resource to communicate purpose.
+Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) } -Tag @{ release = 'GA'; ruleSet = '2021_12'; } {
+    $resources = @(GetTemplateResources);
+
+    if ($resources.Length -eq 0) {
+        return $Assert.Pass();
+    }
+
+    foreach ($resource in $resources) {
+        $Assert.HasField($resource, 'comments');
```
BernieWhite, line 316: We also care if the comment is empty.

tests/PSRule.Rules.Azure.Tests/Azure.Template.Tests.ps1
BernieWhite: At the moment there isn't an easy way to solve this by looking at the template directly. You could potentially do it based on line number using source info since you are calling via $PSRule.GetContent. 

https://github.com/microsoft/PSRule/blob/1a74b5afd1c07053695b470c8559279430374378/src/PSRule/Pipeline/PipelineReciever.cs#L330

Another approach would be to look at the resources. Currently the comments field is removed but it could be emitted. This would give you a resolved resource name. The `_PSRule` meta property automatically gets read by doesn't have a generic dictionary that we can add anything to. You could emit an issue though similar to https://github.com/Azure/PSRule.Rules.Azure/blob/ae50c78797168fa4bd23539e2bafa4a3ed684a4b/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1#L235.

BernieWhite (changes requested):
@ArmaanMcleod Nice. Have a look at the suggestions and let me know what you think.

BernieWhite:
@ArmaanMcleod Another approach to the rule would be to report the number of resources without comments instead of the specific resource name but without producing 4 unnamed reasons.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
ArmaanMcleod: @BernieWhite Nice. Some reason `AllOf` wasn't coming up in my intellisense at the time, seems to show up now.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
ArmaanMcleod: Nice spot, will add this in

ArmaanMcleod:
> @ArmaanMcleod Another approach to the rule would be to report the number of resources without comments instead of the specific resource name but without producing 4 unnamed reasons.

Thanks @BernieWhite for the feedback. Managed to fix the tests by just sorting target name beforehand. 

I've changed the rules to output the template path and number of resources without comments/descriptions. 

I've also created another rule called `Azure.Template.UseDescriptions` to report descriptions for generated templates. Let me know if this is ok.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
```diff
@@ -304,10 +304,64 @@ Rule 'Azure.Template.ValidSecretRef' -Type '.json' -If { (IsParameterFile) } -Ta
     }
 }
 
+# Synopsis: Use comments for each resource in ARM template to communicate purpose.
+Rule 'Azure.Template.UseComments' -Type '.json' -If { (IsTemplateFile) -and !(IsGenerated) } -Tag @{ release = 'GA'; ruleSet = '2021_12'; } {
+    $resources = @(GetTemplateResources);
```
BernieWhite, line 309: What about?

```suggestion
    $resources = @(GetTemplateResources | Where-Object { $Assert.NullOrEmpty($_, "comments").Result });
    $Assert.Count($resources, '.', 0).Reason(
        $LocalizedData.TemplateResourceWithoutComment,
        $TargetObject.FullName,
        $numResourcesWithoutComments
      );
```

BernieWhite (approved):
@ArmaanMcleod Nice. Good to merge. Added a suggestion that might make is slightly simpler.

src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1
ArmaanMcleod: @BernieWhite Nice, this looks much better. I'll do the same for `Azure.Template.UseDescriptions`.

PR #1342, opened 2022-03-28
Fix split on mock properties #1327

### LINKED ISSUE #1327: Index was out of range with split on mock properties
Facing an issue while expanding one of the bicep template that accepts NIC resource id and variables in the template are splitting the resource Id string to get the name of the NIC resource, works fine when deployed however expansion appears to be throwing an error below 

_An error occurred evaluating expression '[reference(resourceId('Microsoft.Network/networkInterfaces', variables('blob_nic_id_segments')[8]), '2020-11-01').ipConfigurations[0].properties.privateIPAddress]' line 0. Index was out of range. Must be non-negative and less than the size of the collection._

Here is a module of bicep that accepts NIC resource id as input and tries to get the NIC resource name using splits (We cannot input NIC name because of other dependencies of nested modules upstream). 

`param config_blob_nic_id string`
`param config_dfs_nic_id string`
`var blob_nic_id_segments = split(config_blob_nic_id, '/')`
`var dfs_nic_id_segments = split(config_dfs_nic_id, '/')`
`resource nic_blob_resource 'Microsoft.Network/networkInterfaces@2020-11-01' existing = {
  name: blob_nic_id_segments[8]
}`
`resource nic_dfs_resource 'Microsoft.Network/networkInterfaces@2020-11-01' existing = {
  name: dfs_nic_id_segments[8] 
}`
`var blob_nic_ip = nic_blob_resource.properties.ipConfigurations[0].properties.privateIPAddress `
`var dfs_nic_ip = nic_dfs_resource.properties.ipConfigurations[0].properties.privateIPAddress`

Note that we are compiling bicep template first so that it generates the json template and then getting it evaluated against PSRule.

BernieWhite:
## PR Summary

- Fixed index was out of range with split on mock properties.

Fixes #1327

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Other code changes**
  - [x] Link to a filed issue

### DIFF
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -14,6 +14,11 @@ See [troubleshooting guide] for a workaround to this issue.
 
 ## Unreleased
 
+What's changed since pre-release v1.14.0-B2203088:
+
+- Bug fixes:
+  - Fixed index was out of range with split on mock properties. [#1327](https://github.com/Azure/PSRule.Rules.Azure/issues/1327)
+
 ## v1.14.0-B2203088 (pre-release)
 
 What's changed since pre-release v1.14.0-B2203066:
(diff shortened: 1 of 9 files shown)


ArmaanMcleod (approved):
@BernieWhite Apologise on the delay with this one, looks greate mate :)

PR #1505, opened 2022-07-04
Fixed transparentDataEncryption property status to state

Dylan-Prins:
## PR Checklist

- [ ] PR has a meaningful title
- [ ] Summarized changes
- [ ] Change is not breaking
- **Rule changes**
  - [ ] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1
@@ -94,7 +94,12 @@ Rule 'Azure.SQL.TDE' -Ref 'AZR-000191' -Type 'Microsoft.Sql/servers/databases' -
         return $Assert.Fail($LocalizedData.SubResourceNotFound, 'Microsoft.Sql/servers/databases/transparentDataEncryption');
     }
     foreach ($config in $configs) {
-        $Assert.HasFieldValue($config, 'Properties.state', 'Enabled');
+        if ($Assert.HasFieldValue($config, 'apiVersion', '2014-04-01').Result) {
+            $Assert.HasFieldValue($config, 'Properties.status', 'Enabled');
+        }
+        else {
+            $Assert.HasFieldValue($config, 'Properties.state', 'Enabled');
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


BernieWhite:
@Dylan-Prins Thanks for the PR. Looks like there is a difference between various API versions or the provider documentation is wrong. The newer use `state` where the older is `status`. If we can verify that `status` was an error, happy to accept you change. Otherwise we should update it to accept either `status` or `state` as `Enabled`.

If the later, to get the tests to pass we need to add an entry for `database-C` that uses the `state` property here: https://github.com/Azure/PSRule.Rules.Azure/blob/ca694e0833548b49cf004f9e30abb156276c55dd/tests/PSRule.Rules.Azure.Tests/Resources.SQL.json#L436-L491

[2014-04-01](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/2014-04-01/servers/databases/transparentdataencryption?tabs=bicep)

[latest](https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/servers/databases/transparentdataencryption?tabs=bicep)

@ms-sambell Did you notice something similar?

Dylan-Prins:
I see! I will add an if statement.
How can I check on API-Version in the if statement?

src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1
```diff
@@ -94,7 +94,7 @@ Rule 'Azure.SQL.TDE' -Ref 'AZR-000191' -Type 'Microsoft.Sql/servers/databases' -
         return $Assert.Fail($LocalizedData.SubResourceNotFound, 'Microsoft.Sql/servers/databases/transparentDataEncryption');
     }
     foreach ($config in $configs) {
-        $Assert.HasFieldValue($config, 'Properties.status', 'Enabled');
+        $Assert.HasFieldValue($config, 'Properties.state', 'Enabled');
```
BernieWhite, line 97: The property `apiVersion` will not exist if exported with _in-flight_ scenarios, but latest API version should be used. We only need to catch the case when older IaC is used.

```suggestion
        if ($Assert.HasFieldValue($config, 'apiVersion', '2014-04-01').Result) {
            $Assert.HasFieldValue($config, 'Properties.status', 'Enabled');
        }
        else {
            $Assert.HasFieldValue($config, 'Properties.state', 'Enabled');
        }
```

BernieWhite (changes requested):
@Dylan-Prins Nice work on the test resource. I think your suggestion is better to conditionally detect based on API version, but there is some specific cases around it.

Based on your suggestion, I am fine with removing `database-C` and just updating `database-B` with `state` as an exports from Azure will be using a newer API version the specific case we are testing should always be `state`.

BernieWhite (approved):
@Dylan-Prins Great, all good to go. I'll update the branch for main and merge. Thanks!

Issue #1610, opened 2022-09-03
Reference to Azure Security Benchmark


BernieWhite:
### Discussed in https://github.com/Azure/PSRule.Rules.Azure/discussions/1244

<div type='discussions-op-text'>

<sup>Originally posted by **BernieWhite** February  8, 2022</sup>
Some rules specifically check security controls defined in [Azure Security Benchmark](https://docs.microsoft.com/security/benchmark/azure/).

ASB provides a reference point to map Azure controls to other security standards, for [example](https://docs.microsoft.com/security/benchmark/azure/security-controls-v3-privileged-access):

CIS Controls v8 ID(s) | NIST SP 800-53 r4 ID(s) | PCI-DSS ID(s) v3.2.1
-- | -- | --
5.4, 6.8 | AC-2, AC-6 | 7.1, 7.2, 8.1
</div>

---

This enhancement aims to map rules from the Security pillar to Azure Security Benchmark.

Rules will be tagged with:

- `Azure.WAF/pillar` which will be set to `Security`

Rules will be labeled with (to support mapping to multiple controls if applicable):

- `Azure.ASB.v3/control` which will be set to the control ID from ASB.

Syntax:

```yaml
Azure.WAF/pillar: 'Security'
Azure.ASB.v3/control: 'XX-N'
```

For example, in YAML:

```yaml
---
# Synopsis: Internet accessible Application Gateways should use protect endpoints with WAF.
apiVersion: github.com/microsoft/PSRule/v1
kind: Rule
metadata:
  name: Azure.AppGw.UseWAF
  ref: AZR-000063
  tags:
    release: 'GA'
    ruleSet: '2020_06'
    Azure.WAF/pillar: 'Security'
  labels:
    Azure.ASB.v3/control: 'NS-6'
spec:
  with:
  - Azure.IsAppGwPublic
  condition:
    field: properties.sku.tier
    in:
    - WAF
    - WAF_v2
```

For example in PowerShell:

```powershell
# Synopsis: App Service should reject TLS versions older then 1.2.
Rule 'Azure.AppService.MinTLS' -Ref 'AZR-000073' -Type 'Microsoft.Web/sites', 'Microsoft.Web/sites/slots' -Tag @{ release = 'GA'; ruleSet = '2020_06'; 'Azure.WAF/pillar' = 'Security'; } -Labels @{ 'Azure.ASB.v3/control' = 'DP-3' } {
[...]

Issue #1109, opened 2021-11-25
Rule refactoring for YAML


BernieWhite:
Most rules are currently written in PowerShell. YAML based rules added in PSRule v1.6.0 provide some advantages including:

- Easier to adopt for non-PowerShell engineers.
- Better performance.

We should relook at existing rules and determine which can be converted to YAML-based syntax.

PR #1611, opened 2022-09-03
Rule updates with metadata and ASB controls #1610

BernieWhite:
## PR Summary

- Updates to rule metadata to progress towards #1610.
- Convert of additional PowerShell rules to YAML towards #1109.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue

### DIFF
--- a/docs/en/baselines/Azure.All.md
+++ b/docs/en/baselines/Azure.All.md
@@ -4,7 +4,7 @@ Includes all Azure rules.
 
 ## Rules
 
-The following rules are included within `Azure.All`. This baseline includes a total of 283 rules.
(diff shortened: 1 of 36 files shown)


ms-sambell (approved):
LGTM - It might be worth documenting the expected categories so it doesn't get out of hand again. Maybe a test to ensure that the category for a role is in the approved list?

BernieWhite:
> LGTM - It might be worth documenting the expected categories so it doesn't get out of hand again. Maybe a test to ensure that the category for a role is in the approved list?

The WAF does change from time to time so the categories change which is why some updates are required. However that might make it clearer and easier for contributors if there was a list of categories for each pillar.

Issue #1618, opened 2022-09-04
[New rule] AKS clusters should use ephemeral ephemeral OS disks which can provide lower read/write latency, along with faster node scaling and cluster upgrades


bengeset96:
# Rule request

## Suggested new rule

AKS clusters should use ephemeral ephemeral OS disks which can provide lower read/write latency, along with faster node scaling and cluster upgrades.

## How the rule should work

Should check for OS disk type is of `Ephemeral` and not `Managed`.

## Additional

Ephemeral is a bit complicated, so the documentation on this rule must be really good. Even when configuring `Ephemeral` it is easy to get validation error since the OS disk must fit in the VM cache. Correct OS disk size is therefore very important.

I have checked and there is no "easy" solution to check cache size on the `osDiskSizeGB`, so good documentation will be key here when it throws error/s to help the customer to figure out why the rule is enforcing something that can give validation errors.

https://docs.microsoft.com/en-us/azure/aks/cluster-configuration#ephemeral-os

bengeset96:
@BernieWhite 

To be closed?

PR #1619, opened 2022-09-04
[New rule] Azure.AKS.EphemeralOSDisk

bengeset96:
## PR Summary

Working on #1618 

Currently working on this rule and have some questions, specially towards the function called `GetSubResources`.

For AKS the OS disk is configured under `$TargetObject.properties.agentPoolProfiles.osDiskType` when using `Microsoft.ContainerService/managedClusters`. On the other hand it seems like when using the child resource `Microsoft.ContainerService/managedClusters/agentPools` it goes under `$TargetObject.properties.osDiskType`.

Based on the work that has been done with the AKS rules earlier, I found out that the global function `GetSubResources` has been used.

image

My question is simply how this function is working (been trying to take a look on it, but it was currently little bit to complicated for me), but how is it able to find a resource type that is not declared in `Microsoft.ContainerService/managedClusters`? As you can see in my commit I am using this function with great success, but I am struggling how it is able to find the resource type that is not explicitly stated in the ARM template.

Does it do some sort of magic and understand that when stating `GetSubResources -ResourceType 'Microsoft.ContainerService/managedClusters/agentPools'` and understand that that is `properties.agentPoolProfiles` in `'Microsoft.ContainerService/managedClusters`?

Since this child resource is not stated in the ARM template unless you specify it explicitly. Seems like MS is handling it for you when using `Microsoft.ContainerService/managedClusters`. 

So the million dollar question is... How does this function do the black magic I am not able to see?

I hope my question is understandable.

## PR Checklist

- [ ] Summarized changes
- [ ] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


BernieWhite:
@bengeset96 The agent pools can either be a sub-resource `Microsoft.ContainerService/managedClusters/agentPools` or configured under the `properties.agentPoolProfiles` property of the AKS cluster resource `Microsoft.ContainerService/managedClusters`. Both are valid infrastructure as code options to create/ configure agent pools.

For example you might deploy a cluster then add an agent pool by using `Microsoft.ContainerService/managedClusters/agentPools`, or it should be contained in separate Bicep module in the same deployment.

When using the `Microsoft.ContainerService/managedClusters/agentPools` sub-resource properties for the agent pool are under the `properties` property like most resourced defined in Azure templates. Where they are direct properties of each array element from the properties.agentPoolProfiles` array.

Does that make sense?

bengeset96:
> @bengeset96 The agent pools can either be a sub-resource `Microsoft.ContainerService/managedClusters/agentPools` or configured under the `properties.agentPoolProfiles` property of the AKS cluster resource `Microsoft.ContainerService/managedClusters`. Both are valid infrastructure as code options to create/ configure agent pools.
> 
> For example you might deploy a cluster then add an agent pool by using `Microsoft.ContainerService/managedClusters/agentPools`, or it could be contained in separate Bicep module in the same deployment.
> 
> When using the `Microsoft.ContainerService/managedClusters/agentPools` sub-resource properties for the agent pool are under the `properties` property like most resourced defined in Azure templates. Where they are direct properties of each array element from the `properties.agentPoolProfiles` array.
> 
> Does that make sense?

I think so, it's maybe my description of what I'm wondering about that is not clear enough.

I am wondering how the global function `GetSubResources` with parameter `-ResourceType 'Microsoft.ContainerService/managedClusters/agentPools'` is able to find the resource on a file where the resource is not declared explictly like `'Microsoft.ContainerService/managedClusters/agentPools` and only `'Microsoft.ContainerService/managedClusters` is to be found on the resource declaration.

This can be found in this example created by Armaan where it's using this function when there is only `'Microsoft.ContainerService/managedClusters'` to be found in the resource declaration.

https://github.com/Azure/PSRule.Rules.Azure/blob/ad6b71556d1d27b7db41f09986c7db71005eb8b5/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1#L250

BernieWhite:
@bengeset96 `GetSubResources` isn't doing anything that special, it mostly just looks for the `resources` property array on a resource. It's defined here https://github.com/Azure/PSRule.Rules.Azure/blob/main/src/PSRule.Rules.Azure/rules/Azure.Common.Rule.ps1#L34-L65

`GetSubResource` just filtered any nested resources found within the `resources` property based on name and type.

BernieWhite:
This block https://github.com/Azure/PSRule.Rules.Azure/blob/ad6b71556d1d27b7db41f09986c7db71005eb8b5/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1#L249-L259 just combines both the agent pools defined in `Properties.agentPoolProfiles` and any defined as sub-resources within the `resources` property.

bengeset96:
@BernieWhite 

Thanks for being so kind and understandable, I will try to search about the topic to get full understanding instead of utilizing your name on explaining "fundamental knowledge".

Anyways, the rule is coming together nicely, but I am now trying to leverage some of the more advanced functionalities of PSRule.

I am really struggling to get the .Reason method to work from the Assert result object. I want to leverage this so I have well-written reason before passing it further to the pipeline.

Could you perhaps assist me with checking what I've done wrong? I also uploaded what I've done in the latest commit. I have followed the same logic as I've seen on other rules and also checked documentation.

```
$filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.AKS.EphemeralOSDisk' }
$ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' })
$ruleResult.Detail

Reason
------
{}
{}
{}
{}
{}
{}
{}
```

BernieWhite:
@bengeset96 Seems to be working for me. You should test from a local build though (using the VSCode task or `./build.ps1`), as if you are pointing to files directly PSRule may not be able to find the `PSRule-rules.psd1`.

Testing manually you can then run:

```powershell
import-Module .\out\modules\PSRule.Rules.Azure\
$results = invoke-psrule -f tests\PSRule.Rules.Azure.Tests\Resources.AKS.json -m PSRule.Rules.Azure -name Azure.AKS.EphemeralOSDisk -Outcome Fail
$results[0].Detail
```

Your usage is mostly fine except that `osDiskType` will be null if the property is not set of the resource which will result in:

```text
Reason
------
{The OS disk is of type {0}.}
```

Instead of a full replaced string.

BernieWhite:
Also specifically about Ephemeral OS disks it's worth calling out this behaviour:

> When a user does not explicitly request managed disks for the OS, AKS will default to ephemeral OS if possible for a given node pool configuration.

> When using ephemeral OS, the OS disk must fit in the VM cache. The sizes for VM cache are available in the [Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series) in parentheses next to IO throughput ("cache size in GiB").

> Using the AKS default VM size [Standard_DS2_v2](https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series) with the default OS disk size of 100GB as an example, this VM size supports ephemeral OS but only has 86GB of cache size. This configuration would default to managed disks if the user does not specify explicitly. If a user explicitly requested ephemeral OS, they would receive a validation error.

bengeset96:
> @bengeset96 Seems to be working for me. You should test from a local build though (using the VSCode task or `./build.ps1`), as if you are pointing to files directly PSRule may not be able to find the `PSRule-rules.psd1`.
> 
> Testing manually you can then run:
> 
> ```powershell
> import-Module .\out\modules\PSRule.Rules.Azure\
> $results = invoke-psrule -f tests\PSRule.Rules.Azure.Tests\Resources.AKS.json -m PSRule.Rules.Azure -name Azure.AKS.EphemeralOSDisk -Outcome Fail
> $results[0].Detail
> ```
> 
> Your usage is mostly fine except that `osDiskType` will be null if the property is not set on the resource which will result in:
> 
> ```
> Reason
> ------
> {The OS disk is of type {0}.}
> ```
> 
> Instead of a fully replaced string.
> 
> Instead you could potentially use `ReasonIf`.

I've been playing a bit with this since I am really struggling with actually being able to test the reason, because always blank.

So, always on the second time I execute the script it throws an error. 

image

Then when rebooting the computer.. it works again. But same occurs after secound try.

Using the inbuilt build task in VS Code is seems to work fine.

image

On the other hand, after the build:

`PS C:\Users\bengeset\PSRule.Rules.Azure> $results = invoke-psrule -f tests\PSRule.Rules.Azure.Tests\Resources.AKS.json -m PSRule.Rules.Azure -name Azure.AKS.EphemeralOSDisk -Outcome Fail
WARNING: Could not find a matching rule. Please check that Path, Name and Tag parameters are correct.
PS C:\Users\bengeset\PSRule.Rules.Azure> `

Any ideas how I can pull this together?

When doing it the other way like

BernieWhite:
@bengeset96 PowerShell get's a little funny about loading multiple .NET assemblies with the same name within the same session. Building via VSCode is a good idea mainly because it uses a new session each time so don't have issues with loading the same assembly.

After the build just be sure to run `Import-Module .\out\modules\PSRule.Rules.Azure` to load the built module into PowerShell. Otherwise you'll end up with PowerShell loading any released PSRule.Rules.Azure modules you have installed locally with `Install-Module` .

PSRule will automatically look for modules included in your session before it tried to load an existing module.

bengeset96:
@BernieWhite 

Do you mind doing a PR review now?

I think it should be ready. Some points from my side.

1. Rule has been configured with `-Level Warning`. This is due to `Ephemeral` will be configured in some cases, without specifying it explicitly, but the rule will tell you it's not configured. It is also therefore mentioned in the rule documentation that this rule might not be 100% accurate and why that is the case in certain scenarios with examples describing why to help customers. I did not include this in the reason message as I found out it can be little bit to messy with so much information there and rather using rule documentation as support. Feel free to correct me.

2. ReasonIf has been used, so there will be different reasons based on if osDiskType has been specified or not.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -238,6 +238,18 @@ Rule 'Azure.AKS.UptimeSLA' -Ref 'AZR-000285' -Type 'Microsoft.ContainerService/m
...
 
+# Synopsis: AKS clusters should use ephemeral OS disks which can provide lower read/write latency, along with faster node scaling and cluster upgrades.
+Rule 'Azure.AKS.EphemeralOSDisk' -Ref 'AZR-000287' -Level Warning -Type 'Microsoft.ContainerService/managedClusters', 'Microsoft.ContainerService/managedClusters/agentPools' -Tag @{ release = 'GA'; ruleSet = '2022_09' } {
+    $agentPools = @(GetAgentPoolProfiles);
+    if ($agentPools.Length -eq 0) {
+        return $Assert.Pass();
+    }
+    foreach ($agentPool in $agentPools) {
+        $Assert.In($agentPool, 'osDiskType', 'Ephemeral').
+        ReasonIf($agentPool.osDiskType, $LocalizedData.AKSEphemeralOSDiskNotConfigured);
```
BernieWhite, line 249: @bengeset96 We may cause some necessary false positives as you've noted in the docs. Thinking more on this rule, I think it matter less that `Ephemeral` is set, and more that `Managed` has not be explicitly set. Since if a preference is not set either way then ephemeral disks will be used if possible. What do you think?

Another way to address the rule would be to build a mapping of ephemeral disk sizes, but I don't think this would be every effective to maintain longer term.

```suggestion
        $Assert.HasDefaultValue($agentPool, 'osDiskType', 'Ephemeral').
        ReasonIf($agentPool.osDiskType, $LocalizedData.AKSEphemeralOSDiskNotConfigured);
```

BernieWhite (changes requested):
@bengeset96 Good progress. I few updates to the documentation to look at. Also maybe `HasDefaultValue` might be an option that is less likely to generate false positives.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
bengeset96: I fully agree. Actually I should have taught that way around it in the first place, this is much better indeed.

Orginially my idea was the build a mapping, but after some tinkering I choose not to go this path. Reason for this was simply due to that AKS teams in organizations often tend to cattle alot and rebuild from scratch time to time, tweak their configuration, and it would have been unfortunate if the mapping already was outdated and did not give accurate results. I just think for now it will lead to management overhead for us and something that would have to be maintained regularly. I feel like `HasDefaultValue` will be a good trade-off.

bengeset96:
@BernieWhite Thanks for the feedback, all suggestions and brainstorming being totally understandable.

Do you mind taking a look now? I have adjusted accordingly.

docs/CHANGELOG-v1.md
```diff
@@ -23,6 +23,11 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
...
 ## Unreleased
 
+- New rules:
+  - AKS:
+    - Check clusters use Ephemeral OS disk by @bengeset96.
+      [#1618](https://github.com/Azure/PSRule.Rules.Azure/issues/1618)
+
 What's changed since pre-release v1.20.0-B0004:
 
 - New rules:
```
BernieWhite, line 33:
```suggestion
What's changed since pre-release v1.20.0-B0004:

- New rules:
  - AKS:
    - Check clusters use Ephemeral OS disk by @bengeset96.
      [#1618](https://github.com/Azure/PSRule.Rules.Azure/issues/1618)
```

docs/en/rules/Azure.AKS.EphemeralOSDisk.md
```diff
@@ -0,0 +1,184 @@
...
+    "vmSize": "[variables('vmSize')]",
+    "osDiskSizeGB": 60,
+    "osType": "Linux",
+    "osDiskType": "Ephemeral",
+    "maxPods": 50,
+    "mode": "User"
+  },
+  "dependsOn": [
+    "[resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName'))]"
+  ]
```
BernieWhite, line 109:
```suggestion
  ]
}
```

docs/en/rules/Azure.AKS.EphemeralOSDisk.md
```diff
@@ -0,0 +1,184 @@
...
+  name: poolName
+  properties: {
+    count: minCount
+    vmSize: vmSize
+    osDiskSizeGB: 60
+    osType: 'Linux'
+    osDiskType: 'Ephemeral'
+    maxPods: 50
+    mode: 'User'
+  }
```
BernieWhite, line 176:
```suggestion
  }
}
```

docs/en/rules/Azure.AKS.EphemeralOSDisk.md
```diff
@@ -0,0 +1,184 @@
...
+  },
+  "properties": {
+    "dnsPrefix": "[parameters('dnsPrefix')]",
+    "agentPoolProfiles": [
+      {
+        "name": "agentpool",
+        "osDiskSizeGB": 60,
+        "count": "[parameters('agentCount')]",
+        "vmSize": "[parameters('agentVMSize')]",
+        "osDiskType": "Ephemeral"
```
BernieWhite, line 68:
```suggestion
        "osDiskType": "Ephemeral",
```

tests/PSRule.Rules.Azure.Tests/Azure.AKS.Tests.ps1
```diff
@@ -1330,5 +1330,41 @@ Describe 'Azure.AKS' -Tag AKS {
...
+            $ruleResult[7].Reason | Should -Not -BeNullOrEmpty;
+            $ruleResult[7].Reason | Should -BeExactly "Path osDiskType: Does not exist.";
+            $ruleResult[8].Reason | Should -Not -BeNullOrEmpty;
+            $ruleResult[8].Reason | Should -BeExactly "The OS disk type 'Managed' should be of type 'Ephemeral'.";
+
+            # Pass
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Pass' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult.Length | Should -Be 3;
+            $ruleResult.TargetName | Should -BeIn 'cluster-C', 'system', 'cluster-L';
```
BernieWhite, line 1367:
```suggestion
            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' });
            $ruleResult | Should -Not -BeNullOrEmpty;
            $ruleResult.Length | Should -Be 2;
            $ruleResult.TargetName | Should -BeIn 'cluster-F', 'cluster-K';

            $ruleResult[0].Reason | Should -BeExactly "The OS disk type 'Managed' should be of type 'Ephemeral'.";
            $ruleResult[1].Reason | Should -BeExactly "The OS disk type 'Managed' should be of type 'Ephemeral'.";

            # Pass
            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Pass' });
            $ruleResult | Should -Not -BeNullOrEmpty;
            $ruleResult.Length | Should -Be 10;
            $ruleResult.TargetName | Should -BeIn 'cluster-A', 'cluster-B', 'cluster-C', 'cluster-D', 'system', 'cluster-L', 'cluster-G', 'cluster-H', 'cluster-I', 'cluster-J';
```

BernieWhite (changes requested):
@bengeset96 Tests are failing, try these updates.

bengeset96:
> @bengeset96 Tests are failing, try these updates.

Yep, I checked the CI, guess who did not update the tests after changing to the `HasDefaultValue` method..

I'm on it.

bengeset96:
@BernieWhite Should be good now, tests passed.

BernieWhite (approved):
@bengeset96 Thanks for your work on this one. All good to merge.

PR #1638, opened 2022-09-12
Added Azure.VMSS.PublicKey

### LINKED ISSUE #1635: Azure VMSS should use SSH keys instead of common credentials to secure virtual machine scale sets against malicious activities.
# Rule request

## Suggested new rule

Azure VMSS should use SSH keys instead of common credentials to secure virtual machine scale sets against malicious activities.

## How the rule should work

Ensure that 'disablePasswordAuthentication` is configured for Linux VMSS's.

## Additional

Per default when creating Azure VMSS's with Linux OS, password authentication is not disabled without specifying explicitly. I don't know why this is the case, most likely due to give an easier resource declaration deployment experience. This should be disabled and rather use SSH keys instead of common credentials. VMSS's are often heavily used for building and deploying code, so should also be secured correctly.

bengeset96:
## PR Summary

Closes #1635 

This PR contains a new rule that checks Linux VMSS to ensure that password authentication is disabled. 

Linux VMSS has the option to use SSH keys instead which is a better way to authenticate.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


bengeset96:
@BernieWhite 

Do you mind taking a look and give some feedback before I add it in the changelog?

Do you think it's worth adding some ARM and Bicep examples and just try to slim down the template ref? Or is current rule documentation sufficient enough?

I will change the rule name as 35 was maximum, so building the module fails.

BernieWhite:
> @BernieWhite
> 
> Do you mind taking a look and give some feedback before I add it in the changelog?
> 
> Do you think it's worth adding some ARM and Bicep examples and just try to slim down the template ref? Or is current rule documentation sufficient enough?
> 
> I will change the rule name as 35 was maximum, so building the module fails.

@bengeset96 Good idea. I like the rule.

Yes, if we can have a bicep / template example code that would be great. Although some rules doesn't currently have this, it's something that we need to clean up so all rules have good examples.

Generally, I just write a `examples-*.bicep` file in `docs/` which can be easily copied and pasted to multiple docs in the future. Also it's easy to build the bicep code into the ARM example.

The VM version is called `Azure.VM.PublicKey` so maybe something similar.

bengeset96:
@BernieWhite 

Review can be done now. I have changed the name to `Azure.VMSS.PublicKey` so it has the "same" pattern as `Azure.VM.PublicKey`. I think the name makes sense.

docs/en/rules/Azure.VMSS.PublicKey.md
```diff
@@ -0,0 +1,213 @@
...
+Use SSH keys instead of common credentials to secure virtual machine scale sets against malicious activities.
+
+## DESCRIPTION
+
+Linux virtual machine scale sets should have password authentication disabled to help with eliminating password-based attacks.
+
+A common tactic observed used by adversaries against customers running Linux Virtual Machines (VMs) in Azure is password-based attacks.
+
+[Azure security baseline for Linux Virtual Machines](https://docs.microsoft.com/security/benchmark/azure/baselines/virtual-machines-linux-security-baseline)
+[Detailed steps: Create and manage SSH keys for authentication to a Linux VM in Azure](https://docs.microsoft.com/azure/virtual-machines/linux/create-ssh-keys-detailed)
```
BernieWhite, line 22: @bengeset96 We can remove these links as they are already in the links section.

src/PSRule.Rules.Azure/rules/Azure.VMSS.Rule.ps1
```diff
@@ -44,4 +44,23 @@ Rule 'Azure.VMSS.ComputerName' -Ref 'AZR-000262' -Type 'Microsoft.Compute/virtua
     Match 'Properties.virtualMachineProfile.osProfile.computerNamePrefix' $matchExpression
 }
 
+# Synopsis: Use SSH keys instead of common credentials to secure virtual machine scale sets against malicious activities.
+Rule 'Azure.VMSS.PublicKey' -Ref 'AZR-000288' -Type 'Microsoft.Compute/virtualMachineScaleSets' -If { IsLinuxVMSS } -Tag @{ release = 'GA'; ruleSet = '2022_09' } {
+    $Assert.In($TargetObject, 'properties.virtualMachineProfile.OsProfile.linuxConfiguration.disablePasswordAuthentication', $True).
+    Reason($LocalizedData.VMSSPublicKey, $TargetObject.name)
```
BernieWhite, line 50: Basically the same, but in some circumstances a different property for the name might be used such as `resourceName` instead of `name`.

```suggestion
    Reason($LocalizedData.VMSSPublicKey, $PSRule.TargetName)
```

src/PSRule.Rules.Azure/rules/Azure.VMSS.Rule.ps1
```diff
@@ -44,4 +44,23 @@ Rule 'Azure.VMSS.ComputerName' -Ref 'AZR-000262' -Type 'Microsoft.Compute/virtua
...
 #endregion Virtual machine scale set
+
+#region Helper functions
+
+function global:IsLinuxVMSS {
+    [CmdletBinding()]
+    [OutputType([System.Boolean])]
+    param ()
+    process {
+        return $Assert.HasField($TargetObject, 'properties.virtualMachineProfile.OsProfile.linuxConfiguration')
```
BernieWhite, line 62:
```suggestion
        return $Assert.HasField($TargetObject, 'properties.virtualMachineProfile.OsProfile.linuxConfiguration').Result
```

tests/PSRule.Rules.Azure.Tests/Azure.VMSS.Tests.ps1
```diff
@@ -233,5 +233,26 @@ Describe 'Azure.VMSS' -Tag 'VMSS' {
...
+        It 'Azure.VMSS.PublicKey' {
+            $dataPath = Join-Path -Path $here -ChildPath 'Resources.VMSS.json';
+            $result = Invoke-PSRule @invokeParams -InputPath $dataPath
+            $filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.VMSS.PublicKey' };
+
+            # Fail
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult.Length | Should -Be 2;
+            $ruleResult.TargetName | Should -BeIn 'vmss-002', 'vmss-003';
```
BernieWhite, line 246: Test case for `vmss-002` and `vmss-003` look the same. Maybe `vmss-003` should not have the `disablePasswordAuthentication` property.

BernieWhite (changes requested):
@bengeset96 Nice work. A few suggested changes, otherwise should be fine.

docs/en/rules/Azure.VMSS.PublicKey.md
bengeset96: Agreed - makes more sense.

BernieWhite (approved):
@bengeset96 Thanks. Nice work. All good to merge.

docs/examples-vmss.bicep
```diff
@@ -0,0 +1,79 @@
+// An example simple Linux VMSS
```
BernieWhite, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

// An example simple Linux VMSS
```

BernieWhite (changes requested):
@bengeset96 Just need to add this for CI to pass.

bengeset96:
> @bengeset96 Just need to add this for CI to pass.

Was not aware, been added now!

BernieWhite (approved):
@bengeset96 All good to merge.

PR #1701, opened 2022-09-22
Improve Azure.VM.PublicKey and Azure.VMSS.PublicKey

### LINKED ISSUE #1704: The way of checking that a VM or VMSS has Linux doesn't cover all cases
The way PSRule currently checks if a VM or VMSS has Linux installed doesn't cover all cases: `storageProfile.osDisk.osType` and `virtualMachineProfile.osProfile.linuxConfiguration` could be undefined in an ARM template

This check is used for the preconditions of Azure.VM.PublicKey and Azure.VMSS.PublicKey

VeraBE:
## PR Summary

Closes #1704 

Improve Azure.VM.PublicKey and Azure.VMSS.PublicKey by checking if the VM(SS)s have Linux in a more reliable way: storageProfile.osDisk.osType and virtualMachineProfile.OsProfile.linuxConfiguration could be undefined in an ARM template

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


src/PSRule.Rules.Azure/rules/Azure.Common.Rule.ps1
```diff
@@ -178,15 +178,47 @@ function global:SupportsHybridUse {
     }
 }
 
-function global:IsLinuxOS {
+function global:IsLinuxOS { # TODO doesn't need to be global
+    [CmdletBinding()]
```
VeraBE, line 182: This function doesn't need to be global but I didn't manage to make it work by declaring it as local

src/PSRule.Rules.Azure/rules/LinuxOffers.Rule.ps1
```diff
@@ -0,0 +1,4618 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+$global:LinuxOffers = @(
```
VeraBE, line 4: This is temporal until we can come up with a good way to keep the information up to date. Let me know if I should add the script I used to generate it somewhere in particular

src/PSRule.Rules.Azure/rules/Azure.Common.Rule.ps1
BernieWhite: The discovery and caching process of rules requires a global scoping. Which is a little different then how a PowerShell script by itself would operate. Functions that are in a local or script scope are not available when the rules execute.

See [Define helper functions](https://microsoft.github.io/PSRule/v2/scenarios/azure-resources/azure-resources/#define-helper-functions).

src/PSRule.Rules.Azure/rules/LinuxOffers.Rule.ps1
BernieWhite: @VeraBE Works for me. Let's talk about how we can get the list of offers from ARM. Maybe from a PowerShell call.

Ultimately if we can generate it as JSON in a CI process that would be awesome. 

This is what we have done previously for provider data exports. 

https://github.com/Azure/PSRule.Rules.Azure/blob/main/scripts/providers.psm1
https://github.com/Azure/PSRule.Rules.Azure/blob/main/.github/workflows/providers.yml

BernieWhite (approved):
@VeraBE Nice work! All good to merge.

PR #1710, opened 2022-09-23
Added Azure.ACR.SoftDelete rule

### LINKED ISSUE #1674: Container registry soft-delete
# Rule request

## Suggested rule change

Currently soft-delete for container registries is a preview feature that keeps deleted artifacts for a soft-delete interval.

Soft-delete allows accidental actions to be recovered to restore a service otherwise relying on the artifacts that have been deleted.

There are a number of limitations with the feature currently. See [preview limitations](https://learn.microsoft.com/azure/container-registry/container-registry-soft-delete-policy#preview-limitations).

As a result, we shouldn't recommend this feature if a retention policy is configured or if we know that the registry is geo-replicated.

## Applies to the following

- Resource type: **Microsoft.ContainerRegistry/registries**

## Additional context

[CARML](https://github.com/Azure/ResourceModules/pull/2037#discussion_r974329513)
[Update](https://azure.microsoft.com/updates/public-preview-soft-delete-in-azure-container-registry/)
[Docs](https://learn.microsoft.com/azure/container-registry/container-registry-soft-delete-policy)
[Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.containerregistry/registries#policies)

bengeset96:
## PR Summary

Added Azure.ACR.SoftDelete rule.

Azure Container Registries should have soft delete policy enabled.

Fixes #1674

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,13 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
     return $Assert.Fail($LocalizedData.ReplicaNotFound);
 }
 
+# Synopsis: Azure Container Registries should have soft delete policy enabled and retention period to 90 days.
+Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000299' -Type 'Microsoft.ContainerRegistry/registries' -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
+    $notGeoReplica = @(GetSubResources -ResourceType 'Microsoft.ContainerRegistry/registries/replications')
+    $Assert.Count($notGeoReplica, '.', 0).Reason($LocalizedData.ACRNotGeoReplica, $TargetObject.name)
```
bengeset96, line 54: @BernieWhite 

I'm little lost here and need some help. Also reason for that I uploaded the drafted product.

What is the best method definition to use to ensure that the array is empty? Looks like all the methods requires the field parameter in the overload defintions. All I want to do is to ensure that the array is of count 0. I found this example used in another rule with the use of `'.'`, that I suppose reflect the object itself? So object, object and count if I'm correct?

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,13 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
     return $Assert.Fail($LocalizedData.ReplicaNotFound);
 }
 
+# Synopsis: Azure Container Registries should have soft delete policy enabled and retention period to 90 days.
+Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000299' -Type 'Microsoft.ContainerRegistry/registries' -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
```
BernieWhite, line 52: Use `AZR-000310` as there is several PRs and recently merged change that clash. Which is the reason for the build error.

```suggestion
Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000310' -Type 'Microsoft.ContainerRegistry/registries' -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
```

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,13 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
...
+    $notGeoReplica = @(GetSubResources -ResourceType 'Microsoft.ContainerRegistry/registries/replications')
+    $Assert.Count($notGeoReplica, '.', 0).Reason($LocalizedData.ACRNotGeoReplica, $TargetObject.name)
+    $Assert.HasDefaultValue($TargetObject, 'properties.policies.retentionPolicy.status', 'disabled').Reason($LocalizedData.ACRNotRetentionPolicy, $TargetObject.name)
```
BernieWhite, line 55: These should be moved to a pre-condition. We want the rule to be skipped if it meets these conditions. To do this you could define an `-If { }` block on the rule and put these in a separate filtering function.

If the `-If { }` block returns `$True` the rule is run. If the `-If { }` block returns `$False` the rule will be skipped.
An example of the `-If` pre-condition.

https://github.com/Azure/PSRule.Rules.Azure/blob/724e0eb11dbc4935165d6d41ee9a3a984ba14e7c/src/PSRule.Rules.Azure/rules/Azure.Storage.Rule.ps1#L77

An example of the filtering function:

https://github.com/Azure/PSRule.Rules.Azure/blob/724e0eb11dbc4935165d6d41ee9a3a984ba14e7c/src/PSRule.Rules.Azure/rules/Azure.Storage.Rule.ps1#L118-L128

Also see this as additional reference: https://microsoft.github.io/PSRule/v2/scenarios/azure-resources/azure-resources/#define-helper-functions

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,13 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
...
+    $Assert.In($TargetObject, 'properties.policies.softDeletePolicy.status', 'enabled').Reason($LocalizedData.ACRSoftDeletePolicy, $TargetObject.name)
+    $Assert.In($TargetObject, 'properties.policies.softDeletePolicy.retentionDays', 90).Reason($LocalizedData.ACRSoftDeletePolicyRetention, $TargetObject.name)
```
BernieWhite, line 57:
```suggestion
    $Assert.HasFieldValue($TargetObject, 'properties.policies.softDeletePolicy.status', 'enabled').Reason($LocalizedData.ACRSoftDeletePolicy, $TargetObject.name)
    $Assert.HasFieldValue($TargetObject, 'properties.policies.softDeletePolicy.retentionDays').Reason($LocalizedData.ACRSoftDeletePolicyRetention, $TargetObject.name)
```

BernieWhite (changes requested):
@bengeset96 That's a good start. We just need to reorganize it slightly to only run against non-replicated registries without a retention policy.

Have a look at my comments below and let me know if you have any further questions.

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
bengeset96: Will look into this. I started to use a pre-condition, but it ended up with the rule not being ran when we want to show customers that their passed or failed the rule. So the rule should be skipped instead?

```function global:GetPreviewLimitations {
    [CmdletBinding()]
    [OutputType([System.Boolean])]
    param ()
    process {
        $notGeoReplica = @(GetSubResources -ResourceType 'Microsoft.ContainerRegistry/registries/replications')
        $notRetentionPolicy = $Assert.HasDefaultValue($TargetObject, 'properties.policies.retentionPolicy.status', 'disabled').Result
        ($notGeoReplica.Count -eq 0) -and ($notRetentionPolicy -eq $true)
    }   
}

File: src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1 (hunk 1 above)
Comment (bengeset96) [reply src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1:54]: @BernieWhite 

I think you didn't see this question from my side, thanks in advance.

Comment (BernieWhite) [reply src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1]: @bengeset96 Yes `.` refers to the object itself which is passed into the first parameter.

**Comment by bengeset96** [comment]
@BernieWhite 

Ready for PR.

Notes from my side

- The rule will only run when geo-replication and retention policy are not configured. This is preview limitation. This has both advantages and disadvantages, but preview limitations are what they are currently. So not all customers can see that the rule passed or failed for now.
- The rule will require refactoring once the limitations are gone, so we have to wait out for that and adjust accordingly.

File: docs/en/rules/Azure.ACR.SoftDelete.md [hunk 5]
```diff
@@ -0,0 +1,123 @@
+---
+severity: Important
+pillar: Reliability
+category: Data management
+resource: Azure Container Registry
```
BernieWhite, line 5: Other resources are under `Container Registry`, let's use that. https://azure.github.io/PSRule.Rules.Azure/en/rules/resource/#container-registry

```suggestion
resource: Container Registry
```

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,25 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
...
+Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000310' -Type 'Microsoft.ContainerRegistry/registries' -If { GetPreviewLimitations } -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
+    $Assert.HasFieldValue($TargetObject, 'properties.policies.softDeletePolicy.status', 'enabled').Reason($LocalizedData.ACRSoftDeletePolicy, $TargetObject.name)
+    $Assert.HasFieldValue($TargetObject, 'properties.policies.softDeletePolicy.retentionDays').Reason($LocalizedData.ACRSoftDeletePolicyRetention, $TargetObject.name)
+}
+
 #endregion Rules
+
+#region Helper functions
+
+function global:GetPreviewLimitations {
```
BernieWhite, line 61:
```suggestion
function global:GetACRSoftDeletePreviewLimitations {
```

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
```diff
@@ -48,4 +48,25 @@ Rule 'Azure.ACR.GeoReplica' -Ref 'AZR-000004' -Type 'Microsoft.ContainerRegistry
     return $Assert.Fail($LocalizedData.ReplicaNotFound);
 }
 
+# Synopsis: Azure Container Registries should have soft delete policy enabled.
+Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000310' -Type 'Microsoft.ContainerRegistry/registries' -If { GetPreviewLimitations } -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
```
BernieWhite, line 52: Let's make this a little more specific so it has less chance of clashing accidentally with a function created in the future.

```suggestion
Rule 'Azure.ACR.SoftDelete' -Ref 'AZR-000310' -Type 'Microsoft.ContainerRegistry/registries' -If { GetACRSoftDeletePreviewLimitations } -Tag @{ release = 'Preview'; ruleSet = '2022_09'; } {
```

BernieWhite (changes requested):
@bengeset96 Great work! Just a few minor tweaks otherwise all good.

src/PSRule.Rules.Azure/rules/Azure.ACR.Rule.ps1
bengeset96: Agreed, this makes sense!

bengeset96:
> @bengeset96 Great work! Just a few minor tweaks otherwise all good.

@BernieWhite 

Done. I think we are ready.

BernieWhite (approved):
Thanks @bengeset96, great work! All good to merge.

PR #1718, opened 2022-09-26
Added Azure.AppConfig.AuditLogs

### LINKED ISSUE #1690: Audit App Configuration stores
# Rule request

## Suggested rule change

Configuration stores provide diagnostics logs that can be configured to capture audit logs. Audit logs are required for determining which identities are reading configuration.

App Configuration uses the `Audit` category for audit logs.

## Applies to the following

- Resource type: **Microsoft.AppConfiguration/configurationStores**

## Additional context

[Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.appconfiguration/configurationstores)

Related rules:

- https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.Automation.AuditLogs/

bengeset96:
## PR Summary

Fixes #1690 

Added Azure.AppConfig.AuditLogs rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


src/PSRule.Rules.Azure/rules/Azure.AppConfig.Rule.ps1
```diff
@@ -0,0 +1,35 @@
...
+        })
+
+        $Assert.Greater($auditLogs, '.', 0).ReasonFrom(
+            'properties.logs',
+            $LocalizedData.AutomationAccountDiagnosticSetting,
+            $setting.name,
+            'Audit',
+            $joinedLogCategoryGroups
+        ).PathPrefix($path)
+    }
```
BernieWhite, line 32: @bengeset96 Great start. Although realizing now there is a bug in `Azure.Automation.AuditLogs`. Which is if multiple diagnostic log settings are configured they all must have auditing enabled, which is not really a requirement.

We want to get to, auditing either through category or category group being configured at least once.

Have a look at https://github.com/Azure/PSRule.Rules.Azure/blob/main/src/PSRule.Rules.Azure/rules/Azure.KeyVault.Rule.ps1#L27-L34 which is much simpler but doesn't handle category groups.

Ultimately we want a combination of both, for this rule and we can use the final code to improve the logic of the other two rules.

bengeset96: @BernieWhite

Yes, actually it looks like one resource can support up to _five(5)_ different diagnostic settings. 

I'm in the tinkering process right now. 

```$logCategoryGroups = 'audit', 'allLogs'
    $auditDiagnosticLogs = @(GetSubResources -ResourceType 'Microsoft.Insights/diagnosticSettings' | ForEach-Object {
            $_.Properties.logs | Where-Object { ($_.category -eq 'Audit' -or $_.categoryGroup -in $logCategoryGroups) -and $_.enabled
        }    
    })
   ```
 If doing so, I'd loose the opportunity with having `ReasonFrom()` for each item in the collection with diagnostic name and `PathPrefix()` for each item in the collection. That is a real bummer for the end-user.
 
If I use the some kind of assertion method inside `foreach` or `ForEach-Object` like previously, I'd end up with the exact same problem as we're dealing with now.
 
Without overengineering it, do you see any feasable solution that comes to your mind? Some kind of looping? When one item in the collection returns "true" for audit configured, then write it to to a variable. Then all items in general is checking for this variable and if count "1" is met on the variable, then Pass() and if not, the ReasonFrom method with details about prefix object path, diagnostics name etc. But still the problem will be, if the collection retuns "true" in the middle or late in the collection, then already other items have written to the ReasonFrom method..

Comment (bengeset96) [reply src/PSRule.Rules.Azure/rules/Azure.AppConfig.Rule.ps1]: @BernieWhite Agreed, this makes sense.

**Comment by bengeset96** [comment]
@BernieWhite Do you mind taking a look on my recent push and see if there is anything additional that we should change?

**Comment by bengeset96** [comment]
Lemme try to finish up the rule documentation so we don't have failing CI.

**Comment by bengeset96** [comment]
@BernieWhite 

Should be possible to review this now, rule documentation has been created and minor tweaks.

File: docs/en/rules/Azure.AppConfig.AuditLogs.md [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+
+## SYNOPSIS
+
+Ensure app configuration store audit diagnostic logs are enabled.
+
+## DESCRIPTION
+
+To capture logs that record interactions with data or the settings of the app configuration store, diagnostic settings must be configured.
+
+When configuring diagnostic settings, enabled one of the following:
```
BernieWhite, line 19:
```suggestion
When configuring diagnostic settings, enable one of the following:
```

BernieWhite (changes requested):
@bengeset96 Nice work! All good from a technical standpoint. Good coverage in the tests. One small typo to correct then all good to merge.

docs/en/rules/Azure.AppConfig.AuditLogs.md
bengeset96: @BernieWhite Fixed.

BernieWhite (approved):
@bengeset96 Great work on this one. Thanks. All good to merge.

PR #1752, opened 2022-10-04
Update exported policy rule name/ref properties with hash

### LINKED ISSUE #1653: Policy key already exists
**Description of the issue**

When policy definitions are assigned multiple times (which may have different parameters) the subsequent occurrences fail.

**Expected behaviour**

We need to:

- [x] Ignore adding cases were the same policy definition is assigned with the same parameters to prevent duplicates.
- [x] Generate a unique `name` and `ref` for each policy rule. We should be able to do this via a hash of the condition/ pre-conditions.

For example:

- Ref: `Policy.e749c2d003da`

**Error output**

```text
Export-AzPolicyAssignmentRuleData: An item with the same key has already been added. Key: /providers/Microsoft.Authorization/policyDefinitions/26a828e1-e88f-464e-bbb3-c134a282b9de
```

**Module in use and version:**

- Module: PSRule.Rules.Azure
- Version: **1.20.0-B0028**

Captured output from `$PSVersionTable`:

```text
Name                           Value
----                           -----
PSVersion                      7.2.6
PSEdition                      Core
GitCommitId                    7.2.6
OS                             Microsoft Windows 10.0.22000
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

**Additional context**

Related to #181

ArmaanMcleod:
## PR Summary

Fixes #1653.

Updated `name` and `ref` properties, which use hash of definitionId + condition + pre-conditon.

Also allowed the prefix to configurable from `AZURE_POLICY_RULE_PREFIX` config option or `Export-AzPolicyAssignmentRuleData -Prefix`.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [ ] Link to a filed issue
- **Other code changes**
  - [x] Link to a filed issue


src/PSRule.Rules.Azure/Data/Policy/Models.cs
```diff
@@ -108,6 +108,16 @@ public PolicyDefinition(string definitionId, string name, string description, JO
...
 
+        /// <summary>
+        /// The unique hash of the condition and pre-condition for the rule.
+        /// </summary>
+        public string HashKey { get; set; }
+
+        /// <summary>
+        /// The prefix for the policy rule.
+        /// </summary>
+        public string Prefix { get; set; }
```
BernieWhite, line 119: Currently we set name in the constructor however I think we can just remove it from the constructor and set it directly as `{prefix}.Policy.{hash}` later when we have a hash constructed instead of introducing two new properties which are for the same purpose.

src/PSRule.Rules.Azure/Configuration/ConfigurationOption.cs
```diff
@@ -63,7 +66,8 @@ public bool Equals(ConfigurationOption other)
                 Tenant == other.Tenant &&
                 ManagementGroup == other.ManagementGroup &&
                 ParameterDefaults == other.ParameterDefaults &&
-                Deployment == other.Deployment;
+                Deployment == other.Deployment &&
+                PolicyRulePrefix.Equals(other.PolicyRulePrefix, StringComparison.OrdinalIgnoreCase);
```
BernieWhite, line 70: This is for object equity `==` is fine.

```suggestion
                PolicyRulePrefix == other.PolicyRulePrefix;
```

BernieWhite (changes requested):
@ArmaanMcleod Nice work! I've added a few suggestions. If you have any questions let me know.

docs/concepts/about_PSRule_Azure_Configuration.md
```diff
@@ -285,9 +286,38 @@ Example:
...
+
+```yaml
+# YAML: Override the prefix of exported policy rules
+  AZURE_POLICY_RULE_PREFIX: 'AzureCustomPrefix'
+```
+
 ## NOTE
 
-An online version of this document is available at https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/concepts/about_PSRule_Azure_Configuration.md.
+An online version of this document is available at <https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/concepts/about_PSRule_Azure_Configuration.md>.
```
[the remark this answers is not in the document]
ArmaanMcleod: @BernieWhite Seems like markdownlint doesn't like bare urls: `MD034/no-bare-urls: Bare URL used`

ArmaanMcleod:
@BernieWhite Have updated the code with the feedback, thanks for catching all of that 🙂.

BernieWhite (approved):
@ArmaanMcleod Great work! All good to merge.

PR #1750, opened 2022-10-04
Update Azure.RSV.ReplicationAlert.md

aavdberg:
Location is read-only in Bicep

## PR Checklist

- [ ] PR has a meaningful title
- [ ] Summarized changes
- [ ] Change is not breaking
- [ ] This PR is ready to merge and is not **Work in Progress**
- **Rule changes**
  - [ ] Unit tests created/ updated
  - [ ] Rule documentation created/ updated
  - [ ] Link to a filed issue
  - [ ] [Change log](https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/CHANGELOG-v1.md) has been updated with change under unreleased section
- **Other code changes**
  - [ ] Unit tests created/ updated
  - [ ] Link to a filed issue
  - [ ] [Change log](https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/CHANGELOG-v1.md) has been updated with change under unreleased section

PR #1757, opened 2022-10-05
Update Azure.RSV.ReplicationAlert.md

aavdberg:
Yesterday I did a PR to remove the location because this is a read-only option, but i think it had to be locale instead of location. So changed it and added explanation about it.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [ ] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


BernieWhite (commented):
@aavdberg Your removal of `location` and justification in #1750 was correct. The `locale` property is valid however need to understand in a little bit more detail the reason for your proposed change. From what I can tell, `locale` is an optional property.

BernieWhite (changes requested):
@aavdberg I need further information on this to progress any further with review.

> Your removal of location and justification in #1750 was correct. The `locale` property is valid however need to understand in a little bit more detail the reason for your proposed change. From what I can tell, `locale` is an optional property.

aavdberg:
> @aavdberg I need further information on this to progress any further with review.
> 
> > Your removal of location and justification in #1750 was correct. The `locale` property is valid however need to understand in a little bit more detail the reason for your proposed change. From what I can tell, `locale` is an optional property.

Good to know for people that you can also specify the language of the notification message that is send. But your right it is optional.

aavdberg:
Hi @BernieWhite did you see my comment?

bengeset96:
@aavdberg 

The default value when not specifying the property is `en-us`.

We want to have the rule documentation simple as possible for the customers, so there aren't any extra momentums added that can confuse or add unnecessary complexity (to a topic that is already complex for many organizations..).

Although I really like that you want to provide users options for getting a tailored experience =)

Lets see what Bernie answers when he has the time.

docs/en/rules/Azure.RSV.ReplicationAlert.md
```diff
@@ -70,6 +72,7 @@ resource testRecoveryServices 'Microsoft.RecoveryServices/vaults/replicationAler
     customEmailAddresses: [
       'example@email.com'
     ]
+    locale: 'en-US'
```
BernieWhite, line 75: Let's add this `locale` property to the ARM Template code as well so that it is the same between both.

docs/en/rules/Azure.RSV.ReplicationAlert.md
```diff
@@ -59,6 +59,8 @@ alert via a Bicep either configure the `sendToOwners` or `CustomerEmailAddress`
 - Set `properties.sendToOwners` to `Send`.
 - Set `properties.customEmailAddresses` to `[ "example@email.com" ]`
 
+with the locale property you can define the locale for the email notification.
```
BernieWhite, line 62: Lets add this to a separate `## NOTES` section so that the required steps are clear.

Add to notes similar to: https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/en/rules/Azure.Defender.SQL.md#notes

BernieWhite (changes requested):
@aavdberg Thanks for the additional clarification. Happy to include the suggestion. Have a look at my comments below.

docs/en/rules/Azure.RSV.ReplicationAlert.md
aavdberg: added the ## NOTES and moved the text under it

aavdberg:
@BernieWhite there is still a changed waiting looks like and then the PR can be closed.

docs/en/rules/Azure.RSV.ReplicationAlert.md
aavdberg: Thanks for adding @BernieWhite

BernieWhite (approved):
@aavdberg Thanks for the update. All good to merge.

Issue #1773, opened 2022-10-09
Check that secure resource properties are set from secure parameters


BernieWhite:
# Rule request

## Suggested rule change

Some resource properties should only be set from secure parameters using the `secureString` or `secureObject` type.

## Applies to the following

The rule applies to the following:

- Resource type: **Microsoft.Resources/deployments**

PR #1774, opened 2022-10-09
Add check for secure values #1773

BernieWhite:
## PR Summary

- Check sensitive resource values use secure parameters.

Working towards #1773.

@VeraBE This is what I'm thinking in terms of addressing the requirements for this check.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.Deployment.SecureValue.md
+++ b/docs/en/rules/Azure.Deployment.SecureValue.md
@@ -1,0 +1,94 @@
+---
+reviewed: 2022-10-10
+severity: Critical
+pillar: Security
+category: Infrastructure provisioning
+resource: Deployment
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.Deployment.SecureValue/
+---
+
+# Use secure resource values
+
+## SYNOPSIS
+
+Use secure parameters for setting properties of resources that contain sensitive information.
+
+## DESCRIPTION
+
+Azure Bicep and Azure Resource Manager (ARM) templates can be used to deploy resources to Azure.
+When deploying Azure resources, sensitive values such as passwords, certificates, and keys should be passed as secure parameters.
+Secure parameters use the `secureString` or `secureObject` type.
+
+Parameters that do not use secure types are recorded in logs and deployment history.
+These values can be retrieved by anyone with access to the deployment history.
+
+## RECOMMENDATION
+
+Consider using secure parameters for sensitive resource properties.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To configure deployments that pass this rule:
+
+- Set the type of parameters used set sensitive resource properties to `secureString` or `secureObject`.
+
+For example:
+
+```json
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "secret": {
+      "type": "secureString"
+    }
+  },
+  "resources": [
+    {
+      "type": "Microsoft.KeyVault/vaults/secrets",
+      "apiVersion": "2022-07-01",
+      "name": "keyvault/good",
+      "properties": {
+        "value": "[parameters('secret')]"
+      }
+    }
+  ]
+}
+```
+
+### Configure with Bicep
+
+To configure deployments that pass this rule:
+
+- Add the `@secure()` attribute on parameters used to set sensitive resource properties.
+
+For example:
+
+```bicep
+@secure()
+param secret string
+
+resource goodSecret 'Microsoft.KeyVault/vaults/secrets@2022-07-01' = {
+  name: 'keyvault/good'
+  properties: {
+    value: secret
+  }
+}
+```
+
+## NOTES
+
+This rule checks the following resource type properties:
+
+- `Microsoft.KeyVault/vaults/secrets`:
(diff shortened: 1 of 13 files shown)


src/PSRule.Rules.Azure/rules/Azure.Deployment.Rule.ps1
```diff
@@ -55,4 +60,52 @@ function global:RecurseDeploymentSensitive {
...
+        });
+        Write-Debug -Message "Secure parameters are: $($secureParameters -join ', ')";
+
+        foreach ($resource in $resources) {
+            if ($resource.type -eq 'Microsoft.Resources/deployments') {
+                RecurseSecureValue -Deployment $resource;
+            }
+            elseif ($resource.type -eq 'Microsoft.Compute/virtualMachineScaleSets') {
+                if ($resource.properties.virtualMachineProfile.osProfile.adminPassword) {
+                    $isSecure = [PSRule.Rules.Azure.Runtime.Helper]::HasValueFromSecureParameter($resource.properties.virtualMachineProfile.osProfile.adminPassword, $secureParameters);
```
VeraBE, line 88: Ohh, this was my main question, how to know from the rule code that the property was set with a parameter, and get information about it, because while debugging I only saw the final replaced value of the parameters under, for example, $TargetObject.properties.virtualMachineProfile.osProfile.adminPassword, so I didn't know how I could even try to extract the parameter name. I suppose I made a mistake debugging and then didn't express myself properly when asking D: There was no need to do all this, I was planning on doing it myself once I understood that. Thank you very much anyways, it looks very good! Is it okay if I add the rest of the properties to check to this same PR?

tests/PSRule.Rules.Azure.Tests/Azure.Deployment.Tests.ps1
```diff
@@ -54,6 +54,27 @@ Describe 'Azure.Deployment' -Tag 'Deployment' {
...
+        It 'Azure.Deployment.SecureValue' {
+            $sourcePath = Join-Path -Path $here -ChildPath 'Tests.Bicep.9.json';
+            $data = Export-AzRuleTemplateData -TemplateFile $sourcePath -PassThru;
+            $result = $data | Invoke-PSRule @invokeParams -Name 'Azure.Deployment.SecureValue';
+            $filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.Deployment.SecureValue' };
+
+            # Fail
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Fail' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            # $ruleResult.Length | Should -Be 1;
```
VeraBE, line 67: Should this comment be removed?

docs/en/rules/Azure.Deployment.SecureValue.md
```diff
@@ -0,0 +1,94 @@
+---
+reviewed: 2022-10-10
+severity: Critical
+pillar: Security
+category: Infrastructure provisioning
+resource: Deployment
```
VeraBE, line 6: The rule should apply to all resources in the template, not only to the ones under one of type Microsoft.Resources/deployments, why did you make that restriction?

src/PSRule.Rules.Azure/rules/Azure.Deployment.Rule.ps1
VeraBE, line 88: We should probably avoid flagging the property when it was set using keyVault or listKeys, right? Looking for that placeholder value you told me about

tests/PSRule.Rules.Azure.Tests/Azure.Deployment.Tests.ps1
BernieWhite: @verabe I think this one should be uncommented and updated.

docs/en/rules/Azure.Deployment.SecureValue.md
BernieWhite: @verabe This has no bearing on the rule execution just documentation grouping. But we are trying to say this affects all deployments without being specific to template or bicep.

Deployments by there nature are mostly unexpanded. 

I am open to thoughts if it is more or less confusing then all resources. Technically they are both correct. All resources may be less confusing to general consumers.

src/PSRule.Rules.Azure/rules/Azure.Deployment.Rule.ps1
BernieWhite: @verabe You are correct. Good catch. Do you want to have a go updating the validator?

VeraBE: Sure

VeraBE:
I opened a PR with the changes we discussed in the comments: https://github.com/BernieWhite/PSRule.Rules.Azure/pull/169

BernieWhite:
> I opened a PR with the changes we discussed in the comments: [BernieWhite#169](https://github.com/BernieWhite/PSRule.Rules.Azure/pull/169)

@VeraBE Double checking, your updates catch all the known use cases from what I can see as previously discussed, so we should be good to go. Are we good to merge from your perspective?

VeraBE:
@BernieWhite Looks good to merge for me too! Should #1773 be linked again to this PR?

PR #1821, opened 2022-11-02
Added Azure.APIM.MinAPIVersion

### LINKED ISSUE #1819: API Management should minimum use API version 2021-08-01 or newer due to prior versions retiring
# Rule request

## Suggested rule change

API Management should minimum use API version 2021-08-01 or newer due to prior versions retiring.

**On 30 September 2023, all API versions prior to 2021-08-01 will be retired and API calls using those API versions will fail. This means you'll no longer be able to create or manage your API Management services using your existing templates, tools, scripts, and programs until they've been updated. Data operations (such as accessing the APIs or Products configured on Azure API Management) will be unaffected by this update, including after 30 September 2023.**

With implementing a rule on this one, we can help customers with catching this early and help with adjusting accordingly.

`Operational excellence` pillar suits this rule.

## Applies to the following

- Resource type: **[Microsoft.ApiManagement/service]**

## Additional context

[API versions being retired for Azure API Management](https://azure.microsoft.com/updates/api-versions-being-retired-for-azure-api-management)

bengeset96:
## PR Summary

Fixes #1819 

Added Azure.APIM.MinAPIVersion to limit control plane API calls to api management instances with version '2021-08-01' or newer.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


bengeset96:
@BernieWhite 

I need some help and input from you.

- Originally I taught that this was only for the parent resource `Microsoft.ApiManagement/service`, but after reading the documentation again I'm starting to wonder if this is also applicable for child resources? What do you think?
- Can you check if I have from a technical perspective taught out this correctly? `.apiVersion` seems to be an dedicated property for JSON ARM, but when building the Bicep it's under resources. How will PSRule treat this when using `TargetObject.apiVersion`?

```
 "resources": [
    {
      "type": "Microsoft.ApiManagement/service",
      "apiVersion": "2021-12-01-preview",
      "name": "[parameters('apiManagementServiceName')]",
      "location": "[parameters('location')]",
      "sku": {
        "name": "[parameters('sku')]",
        "capacity": "[parameters('skuCount')]"
      },
      "properties": {
        "publisherEmail": "[parameters('publisherEmail')]",
        "publisherName": "[parameters('publisherName')]"
      }
    }
  ]

File: src/PSRule.Rules.Azure/rules/Azure.APIM.Rule.ps1 [hunk 1]
```diff
@@ -244,6 +244,19 @@ Rule 'Azure.APIM.AvailabilityZone' -Ref 'AZR-000052' -Type 'Microsoft.ApiManagem
...
+Rule 'Azure.APIM.MinAPIVersion' -Ref 'AZR-000320' -Type 'Microsoft.ApiManagement/service' -Tag @{ release = 'GA'; ruleSet = '2022_12'; } {
+    [Datetime]$version = $TargetObject.apiVersion -replace '-preview', ''
+    [Datetime]$minimumVersion = '2021-08-01'
+
+    if ($version -ge $minimumVersion) {
+        $Assert.Pass()
+    }
+    else {
+        $Assert.Fail().Reason($LocalizedData.APIMVersionDeprecated, $PSRule.TargetName, $TargetObject.apiVersion)
+    }
```
BernieWhite, line 257: This can be simplified to just, the following. However my understand is that it is actually more to do with a different property `properties.apiVersionConstraint.minApiVersion` which is an optional property. 

I would propose some updates to the rule:

- If `apiVersion` exists then it should be greater than or equal to the minimum. If it don't exist then it must be an in-flight resource so it using the latest resource provider anyway.
- If `properties.apiVersionConstraint.minApiVersion` does not exist, fail.
- If `properties.apiVersionConstraint.minApiVersion` exists then it must be greater or equal to the minimum.
- Lets configure the minimum version as a configuration parameter. A lot of features such as soft-delete and others are based on the API version used to call rest endpoints. So we can expect that the minimum version may need to be updated in the future or customers may want to set a minimum to ensure features are available. This may require updating the docs and reasons slighting to still call out that a minimum version will not be supported but also that the rule failed because the minimum version of X was not met.

https://learn.microsoft.com/en-us/azure/templates/microsoft.apimanagement/service?pivots=deployment-language-bicep#apiversionconstraint

```suggestion
    $Assert.Create($version -ge $minimumVersion, $LocalizedData.APIMVersionDeprecated, $PSRule.TargetName, $TargetObject.apiVersion).PathPrefix('apiVersion')
```

BernieWhite (changes requested):
Thanks @bengeset96. I'd like to take this rule in a slightly different direction if that's ok. I think that should make sense when you reason my comments below but let me know if there are any questions.

BernieWhite:
> @BernieWhite
> 
> I need some help and input from you.
> 
> * Originally I taught that this was only for the parent resource `Microsoft.ApiManagement/service`, but after reading the documentation again I'm starting to wonder if this is also applicable for child resources? What do you think?

I think targeting `Microsoft.ApiManagement/service` is fine since this should mostly be related to the `properties.apiVersionConstraint.minApiVersion` property.

src/PSRule.Rules.Azure/rules/Azure.APIM.Rule.ps1
bengeset96: Good feedback. Changes will be done.

 Could you please explain me what `configuration parameter` would mean in this particular case? I tried to search the `PSRule` and `PSRule for Azure` documentation, but didn't find anything. Do you have an example?

BernieWhite: @bengeset96 Sure. Maybe configuration parameter is unclear, but I mean using the `$Configuration` to set a configuration option that has a default value but can be overridden.

Such as `AZURE_POLICY_WAIVER_MAX_EXPIRY` defined by https://github.com/Azure/PSRule.Rules.Azure/blob/183637188e992a57946da5c83882dde890a78c5d/src/PSRule.Rules.Azure/rules/Azure.Policy.Rule.ps1#L33-L35.

Docs for [configuration](https://microsoft.github.io/PSRule/v2/concepts/PSRule/en-US/about_PSRule_Variables/#configuration).

bengeset96:
@BernieWhite 

Could you take a look on the newest changes? I've spent the day to refactor the rule quite a bit and think we're close on the expected behavior.

I'll update test cases and rule documentation once we have alligned on the rule design.

BernieWhite:
> @BernieWhite
> 
> Could you take a look on the newest changes? I've spent the day to refactor the rule quite a bit and think we're close on the expected behavior.
> 
> I'll update test cases and rule documentation once we have alligned on the rule design.

The direction is good. A few suggestions would be:

- I don't think we need to include the name of the APIM instance in the reason. As this is already part of the rule output.
- I don't think we need to specifically call out deprecation in the reason. The reason the rule failed is because it was less than the minimum version. We should call deprecation and minimum supported version in the rule recommendation docs.
- I think we should also compare the API version of the deployment to `$Configuration.AZURE_APIM_MIN_API_VERSION` as well so they are aligned to the same API version date.

bengeset96:
> > @BernieWhite
> > Could you take a look on the newest changes? I've spent the day to refactor the rule quite a bit and think we're close on the expected behavior.
> > I'll update test cases and rule documentation once we have alligned on the rule design.
> 
> The direction is good. A few suggestions would be:
> 
> * I don't think we need to include the name of the APIM instance in the reason. As this is already part of the rule output.
> * I don't think we need to specifically call out deprecation in the reason. The reason the rule failed is because it was less than the minimum version. We should call deprecation and minimum supported version in the rule recommendation docs.
> * I think we should also compare the API version of the deployment to `$Configuration.AZURE_APIM_MIN_API_VERSION` as well so they are aligned to the same API version date.

Great feedback. Could you take a look on the latest changes and see if I did understand you correctly and if you see anything additional we can make even better/more specific?

src/PSRule.Rules.Azure/rules/Azure.APIM.Rule.ps1
```diff
@@ -244,6 +244,28 @@ Rule 'Azure.APIM.AvailabilityZone' -Ref 'AZR-000052' -Type 'Microsoft.ApiManagem
...
+        $Assert.Create($versionConfigured -ge $minimumVersionExpected, $LocalizedData.APIMApiVersionMin, $TargetObject.apiVersion,
+        $Configuration.AZURE_APIM_MIN_API_VERSION).PathPrefix('apiVersion')
+    }
+    if ($TargetObject.properties.apiVersionConstraint.minApiVersion) {
+        [datetime]$minApiVersionConfigured = $TargetObject.properties.apiVersionConstraint.minApiVersion -replace '-preview', ''
+        [datetime]$minApiVersionExpected = $Configuration.AZURE_APIM_MIN_API_VERSION -replace '-preview', ''
+
+        $Assert.Create($minApiVersionConfigured -ge $minApiVersionExpected, $LocalizedData.APIMApiVersionConstraintMinApiVersion,
+        $TargetObject.properties.apiVersionConstraint.minApiVersion, $Configuration.AZURE_APIM_MIN_API_VERSION).
+        PathPrefix('apiVersionConstraint.minApiVersion')
```
BernieWhite, line 262:
```suggestion
        PathPrefix('properties.apiVersionConstraint.minApiVersion')
```

BernieWhite (changes requested):
@bengeset96 Great work! Just a few tweaks and then all good to merge.

bengeset96:
@BernieWhite 

Latest suggestions has been added.

Thanks for all the great feedback on this one, learned quite much from this PR!

BernieWhite (approved):
Thanks @bengeset96 looks great. All good to merge.

PR #1923, opened 2022-12-01
Added Azure.MariaDB.FirewallRuleCount

### LINKED ISSUE #1855: Azure Database for MariaDB servers should restrict an excessive number of firewall rules
# Rule request

## Suggested rule change

Azure Database for MariaDB servers should restrict an excessive number of firewall rules.

Security pillar for this one.

## Applies to the following

- Resource type: **[Microsoft.DBforMariaDB/servers]**

## Additional context

[Azure Database for MariaDB server firewall rules](https://learn.microsoft.com/azure/mariadb/concepts-firewall-rules)
[Template reference](https://learn.microsoft.com/azure/templates/microsoft.dbformariadb/servers)

BenjaminEngeset:
## PR Summary

Fixes #1855

Added Azure.MariaDB.FirewallRuleCount rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.MariaDB.FirewallRuleCount.md
+++ b/docs/en/rules/Azure.MariaDB.FirewallRuleCount.md
@@ -20,11 +20,16 @@ Server-level firewall rules apply to all databases on the Azure Database for Mar
 
 ## RECOMMENDATION
 
-Review the amount of Azure for MariaDB server firewall rules and consider to reduce the amount where it is possible.
+Review the number of Azure for MariaDB server firewall rules configured.
+Consider to removing rules that are no longer needed.
+
+## NOTES
+
+This rule fails when the number of configured firewall rules exceeds ten (10).
 
 ## LINKS
 
 - [Network security and containment](https://learn.microsoft.com/azure/architecture/framework/security/design-network)
 - [Azure Database for MariaDB server firewall rules](https://learn.microsoft.com/azure/mariadb/concepts-firewall-rules)
 - [Create and manage Azure Database for MariaDB firewall rules by using the Azure portal](https://learn.microsoft.com/azure/mariadb/howto-manage-firewall-portal)
-- [Azure template reference](https://learn.microsoft.com/azure/templates/microsoft.dbformariadb/servers/firewallrules)
+- [Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.dbformariadb/servers/firewallrules)
--- a/docs/en/rules/Azure.MariaDB.FirewallRuleCount.md
+++ b/docs/en/rules/Azure.MariaDB.FirewallRuleCount.md
@@ -1,0 +1,30 @@
+---
+severity: Awareness
+pillar: Security
+category: Network security and containment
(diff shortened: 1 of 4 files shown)


BenjaminEngeset:
Hey @BernieWhite 

- Could you tell me what you're thinking about using this logic? The rule is not bulletproof as it only runs against child resources declared within the parent resource, but from my experience most customers declare child resources within the parent resource anyways.
- I'm waiting for the merge on `Azure.MariaDB.AllowAzureAccess` that will help with good test cases.
- Should we add ARM and Bicep code examples for this one? Where we declare some firewall rules on a MariaDB server?

docs/en/rules/Azure.MariaDB.FirewallRuleCount.md
```diff
@@ -0,0 +1,30 @@
...
+
+## DESCRIPTION
+
+Typically the number of firewall rules required is minimal, with management connectivity from on-premises and cloud application connectivity.
+
+Server-level firewall rules apply to all databases on the Azure Database for MariaDB server.
+
+## RECOMMENDATION
+
+Review Azure for MariaDB server firewall rules.
```
BernieWhite, line 23: I think we need a clearer message here about why the rule failed and what a customer should do.

BernieWhite (commented):
@BenjaminEngeset I have a few comments below but it looks good so far.

For counting the number of firewall rules for the server, this logic looks good.
PSRule will automatically nest sub-resource into the parent resource if they exist in the same deployment/ template when expanding a template. So declaring sub-resources outside of the parent resource is not a concern.

You could write a template with only one firewall rule in it but since we can't make an assumption about how many rules already exist; it's an edge case we don't need to solve in this rule.

BenjaminEngeset:
> @BenjaminEngeset I have a few comments below but it looks good so far.
> 
> > Could you tell me what you're thinking about using this logic? The rule is not bulletproof as it only runs against child resources declared within the parent resource, but from my experience most customers declare child resources within the parent resource anyways.
> 
> For counting the number of firewall rules for the server, this logic looks good. PSRule will automatically nest sub-resource into the parent resource if they exist in the same deployment/ template when expanding a template. So declaring sub-resources outside of the parent resource is not a concern.
> 
> You could write a template with only one firewall rule in it but since we can't make an assumption about how many rules already exist; it's an edge case we don't need to solve in this rule.
> 
> > Should we add ARM and Bicep code examples for this one? Where we declare some firewall rules on a MariaDB server?
> 
> Because this rule is about the number of firewall rules and reducing them, I don't think we need to an example code.

Thanks @BernieWhite. I had no idea that PSRule was that powerful in terms of automatically nesting sub-resources that are declared outside the parent resource, always taught they would been treated as standalones then.

BenjaminEngeset:
@BernieWhite 

Check the latest updates and let me know you're thinking.

BernieWhite (approved):

PR #1927, opened 2022-12-04
Added Azure.VM.AMA

### LINKED ISSUE #1868: Azure Monitor agent should be installed on virtual machines
# Rule request

## Suggested rule change

Azure Monitor agent should be installed on virtual machines.

Defender for Cloud collects data from your Azure virtual machines (VMs) to monitor for security vulnerabilities and threats. Data is collected using the Azure Monitor Agent.

Security pillar for this one.

## Applies to the following

- Resource type: **[Microsoft.Compute/virtualMachines]**

## Additional context

[Deploy the Azure Monitor Agent to protect your servers with Microsoft Defender for Cloud](https://learn.microsoft.com/en-us/azure/defender-for-cloud/auto-deploy-azure-monitoring-agent)
[Template reference](https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachines)

BenjaminEngeset:
## PR Summary

Fixes #1868

Added Azure.VM.AMA rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/rules/Azure.VM.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.VM.Rule.ps1
@@ -279,6 +279,20 @@ Rule 'Azure.VM.SQLServerDisk' -Ref 'AZR-000324' -Type 'Microsoft.Compute/virtual
 
 #endregion IaaS SQL Server disks
 
+#region Azure Monitor Agent
+
+# Synopsis: Use Azure Monitor Agent for collecting monitoring data.
+Rule 'Azure.VM.AMA' -Ref 'AZR-000344' -Type 'Microsoft.Compute/virtualMachines' -Tag @{ release = 'Preview'; ruleSet = '2022_12'; } {
+    $amaTypes = @('AzureMonitorWindowsAgent', 'AzureMonitorLinuxAgent')
+    $extensions = @(GetSubResources -ResourceType 'Microsoft.Compute/virtualMachines/extensions' |
+        Where-Object { $_.properties.publisher -eq 'Microsoft.Azure.Monitor' -or $_.properties.type -in $amaTypes })
+    
+    $Assert.GreaterOrEqual($extensions, '.', 1).
+    Reason($LocalizedData.VMAzureMonitorAgent).PathPrefix('resources')
+}
+
+#endregion Azure Monitor Agent
+
 #region Helper functions
 
 function global:HasPublisherMicrosoftSQLServer {
--- a/src/PSRule.Rules.Azure/rules/Azure.VM.Rule.ps1
+++ b/src/PSRule.Rules.Azure/rules/Azure.VM.Rule.ps1
@@ -282,7 +282,7 @@ Rule 'Azure.VM.SQLServerDisk' -Ref 'AZR-000324' -Type 'Microsoft.Compute/virtual
 #region Azure Monitor Agent
 
 # Synopsis: Use Azure Monitor Agent for collecting monitoring data.
-Rule 'Azure.VM.AMA' -Ref 'AZR-000344' -Type 'Microsoft.Compute/virtualMachines' -Tag @{ release = 'Preview'; ruleSet = '2022_12'; } {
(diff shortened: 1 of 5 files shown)


BenjaminEngeset:
@BernieWhite 

Could you brainstorm on this one with me. 

It all started with #1868 

Currently there is no proper way of configuring this programmatically. The only real option seems to be [Auto-deploy Azure Monitor Agent](https://learn.microsoft.com/en-us/azure/defender-for-cloud/auto-deploy-azure-monitoring-agent). I've requested the docs team to update accordingly. I feel it's a bad idea to point customers to manually configure each virtual machine from the portal. With this option also `Azure Security Agent (ASA) extension` is installed which is needed for Defender for Cloud and specific data collection rules and assignments. These collection rules are not available programmatically, only the basic ones.

Hence, the change in the rule. The rule instead focus towards customers installing the agent for general usage for collecting monitoring data.

It will then be supported by additional rules like `Azure.VM.AMADataCollectionRule/Azure.VM.AMADCR` and `Azure.VM.AMADataCollectionRuleAssignment/Azure.VM.AMADCRAssignment` that checks for that at least one rule is configured and that at lest one assignment is configured.

When documentation and steps are available programatically for the Defender for Cloud part, we introduce an additionally rule like `Azure.AMADefender` that checks for that the agent is installed, ASA extension is installed, correct data collection rules and assignments are configured.

BernieWhite:
> 

@BenjaminEngeset Having a look at this the AMA is deployed via Azure Policy via an assignment _Default Defender for Cloud provisioning Azure Monitor agent_. My guess is there won't will be an auto provisioning configuration similar to previous Log Analytics Agent configuration because Microsoft Defender for Cloud is mostly already configured via policy and `Microsoft.Security/autoProvisioningSettings` was more an artifact of before policy was the primary mechanism to configure MDfC.

So making the rule aligned to OpEx is a good approach. Not entirely sure on the DCR rules, but let's have a look at those in a separate PR. Either way it's still a good rule to have.

src/PSRule.Rules.Azure/rules/Azure.VM.Rule.ps1
```diff
@@ -279,6 +279,20 @@ Rule 'Azure.VM.SQLServerDisk' -Ref 'AZR-000324' -Type 'Microsoft.Compute/virtual
 
 #endregion IaaS SQL Server disks
 
+#region Azure Monitor Agent
+
+# Synopsis: Use Azure Monitor Agent for collecting monitoring data.
+Rule 'Azure.VM.AMA' -Ref 'AZR-000344' -Type 'Microsoft.Compute/virtualMachines' -Tag @{ release = 'Preview'; ruleSet = '2022_12'; } {
```
BernieWhite, line 285: In the context of general monitoring AMA is generally available so we don't need to mark it as preview.

```suggestion
Rule 'Azure.VM.AMA' -Ref 'AZR-000344' -Type 'Microsoft.Compute/virtualMachines' -Tag @{ release = 'GA'; ruleSet = '2022_12'; } {
```

BernieWhite (changes requested):
@BenjaminEngeset Looks great. One minor change.

BenjaminEngeset:
> > 
> 
> @BenjaminEngeset Having a look at this the AMA is deployed via Azure Policy via an assignment _Default Defender for Cloud provisioning Azure Monitor agent_. My guess is there won't will be an auto provisioning configuration similar to previous Log Analytics Agent configuration because Microsoft Defender for Cloud is mostly already configured via policy and `Microsoft.Security/autoProvisioningSettings` was more an artifact of before policy was the primary mechanism to configure MDfC.
> 
> So making the rule aligned to OpEx is a good approach. Not entirely sure on the DCR rules, but let's have a look at those in a separate PR. Either way it's still a good rule to have.

Thanks for the great feedback @BernieWhite and information.

BenjaminEngeset:
@BernieWhite 

Ready for review.

BernieWhite (approved):
Great work @BenjaminEngeset. All good to merge.

PR #2017, opened 2023-02-03
VMs should be fully deallocated

### LINKED ISSUE #88: VMs stopped but not deallocated
# Rule request

## Suggested rule change

VMs that are stopped but not deallocated are still charged for compute. In general virtual machines should be deallocated when not in use.

Edge cases can include domain controllers and VMs for imaging.

## Applies to the following

- Resource type: **Microsoft.Compute/virtualMachines**

dcrreynolds:
## PR Summary

Address issue #88 by gathering VM instance data and adding a new test. I wasn't sure about some things like ruleset and ref, please let me know if I need to change them. 

Fixes #88

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


src/PSRule.Rules.Azure/rules/Azure.VM.Rule.yaml
```diff
@@ -110,6 +110,24 @@ spec:
     field: properties.dnsSettings.dnsServers
     hasValue: false
 
+---
+# Synopsis: VMs should be deallocated instead of stopped.
+apiVersion: github.com/microsoft/PSRule/v1
+kind: Rule
+metadata:
+  name: Azure.VM.ShouldNotBeStopped
+  ref: AZR-123346
```
BernieWhite, line 119: Use the next available ref based on the bottom of this table. https://azure.github.io/PSRule.Rules.Azure/en/rules/

```suggestion
  ref: AZR-000351
```

src/PSRule.Rules.Azure/rules/Azure.VM.Rule.yaml
```diff
@@ -110,6 +110,24 @@ spec:
...
+  tags:
+    release: 'GA'
+    ruleSet: '2020_06'
+    Azure.WAF/pillar: Operational Excellence
+spec:
+  type:
+  - Microsoft.Compute/virtualMachines
+  condition:
```
BernieWhite, line 127: Let's add a selector to only apply to in-flight cases, otherwise this rule may trigger for infrastructure as code which will never has tis value set.

```suggestion
  with:
  - PSRule.Rules.Azure\Azure.Resource.Export
  condition:
```

src/PSRule.Rules.Azure/Pipeline/Export/ResourceExportVisitor.cs
```diff
@@ -281,6 +281,10 @@ private static async Task<bool> VisitVirtualMachine(ResourceContext context, JOb
                         AddSubResource(resource, await GetResource(context, id, APIVERSION_2022_07_01));
                 }
             }
+
+            var response = await context.GetAsync($"{resourceId}/instanceView", APIVERSION_2021_11_01);
+            resource.Add("PowerState", (string)response["statuses"][1]["code"]);
```
BernieWhite, line 286: Will the status we are after always be at index `1`?

I think we should select first or default with `code` starting with `PowerState/` regardless of index, because it could change at any point.

Something like:

```suggestion
            resource.Add("PowerState", response["statuses"].FirstOrDefault(status => status["code"].Value<string>().StartsWith("PowerState/"))["code"].Value<string>());
```

docs/en/rules/Azure.VM.ShouldNotBeStopped.md
```diff
@@ -0,0 +1,27 @@
...
+Azure Virtual Machines in a stopped state are still billed hourly for compute usage. Therefor VMs should generally be in a deallocated or running state.
+
+## RECOMMENDATION
+
+Consider fully deallocating VMs instead of stopping VMs.
+
+## NOTES
+
+## LINKS
+
```
BernieWhite, line 26: Add WAF reference.

```suggestion
## LINKS

- [Shut down underutilized instances](https://learn.microsoft.com/azure/architecture/framework/cost/optimize-vm#shut-down-underutilized-instances)
```

BernieWhite (changes requested):
@dcrreynolds Nice work! Thanks for the PR.  I've add a few suggestions to look at, but most can just be merged. If you have any questions please let me know.

dcrreynolds:
Thanks for the really thorough review! I think I got them all. All tests are passing, and it is working as expected against some VMs I deployed.

BernieWhite (approved):
@dcrreynolds Looks great! All good to merge.

PR #2099, opened 2023-03-20
Added Azure.ContainerApp.ExternalIngress

### LINKED ISSUE #2098: Container Apps should disable external network access
### Existing rule

_No response_

### Suggested rule

Disable external network access for Container Apps by enforcing internal-only ingress. This will ensure inbound communication for Container Apps is limited to callers within the Container Apps environment.

### Pillar

Security

### Additional context

- [Ingress configuration](https://learn.microsoft.com/azure/container-apps/ingress?tabs=bash)
- [Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.app/containerapps#ingress)

BenjaminEngeset:
## PR Summary

Fixes #2098 

Added Azure.ContainerApp.ExternalAccess rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.ContainerApp.ExternalAccess.md
+++ b/docs/en/rules/Azure.ContainerApp.ExternalAccess.md
@@ -1,0 +1,101 @@
+---
+severity: Important
+pillar: Security
+category: Network security and containment
+resource: Container App
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.ContainerApp.ExternalAccess/
+---
+
+# Disable external access
+
+## SYNOPSIS
+
+Ensure inbound communication for Container Apps is limited to callers within the Container Apps environment.
+
+## DESCRIPTION
+
+Container apps allows you to expose your container app to the public web, to your VNET, or to other container apps within your environment by enabling ingress.
+
+Container apps by default will automatically by default have visibility within app environment only if ingress is enabled.
+This secure by default behaviour can be overriden by enabling visibility from internet or VNET, depending on app environment endpoint configured.
+
+Disable external network access to container apps by enforcing internal-only ingress. This will ensure inbound communication for container apps is limited to callers within the container apps environment.
+
+## RECOMMENDATION
+
+Consider enforcing internal-only ingress.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To deploy Container Apps that pass this rule:
+
+- Set `properties.configuration.ingress.external` to `false`.
+
+For example:
+
+```json
+{
+  "type": "Microsoft.App/containerApps",
+  "apiVersion": "2022-10-01",
+  "name": "[parameters('appName')]",
+  "location": "[parameters('location')]",
+  "identity": {
+    "type": "SystemAssigned",
+    "userAssignedIdentities": {}
+  },
+  "properties": {
+        "kubeEnvironmentId": "[resourceId('Microsoft.Web/kubeEnvironments', parameters('envName'))]",
+        "template": {
+            "revisionSuffix": "",
+            "containers": "[variables('containers')]"
+        },
+        "configuration": {
+            "ingress": {
+                "external": false
+            }
+        }
+    },
+    "dependsOn": [
+        "[resourceId('Microsoft.Web/kubeEnvironments', parameters('envName'))]"
+    ]
+}
+```
+
+### Configure with Bicep
+
+To deploy Container Apps that pass this rule:
+
+- Set `properties.configuration.ingress.external` to `false`.
+
+For example:
+
+```bicep
+resource containerApp 'Microsoft.App/containerApps@2022-10-01' = {
+  name: appName
+  location: location
+  identity: {
+    type: 'SystemAssigned'
+    userAssignedIdentities: {}
+  }
+   properties: {
+    kubeEnvironmentId: kubeEnv.id
+    template: {
+      revisionSuffix: ''
+      containers: containers
+    }
+    configuration: {
+      ingress: {
+        external: false
+      }
+    }
+  }
+}
+```
+
+## LINKS
+
+- [UNetworking architecture in Azure Container Apps](https://learn.microsoft.com/azure/container-apps/networking)
+- [Set up HTTPS or TCP ingress in Azure Container Apps](https://learn.microsoft.com/azure/container-apps/ingress)
+- [Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.app/containerapps#ingress)
--- a/docs/en/rules/Azure.ContainerApp.ExternalAccess.md
+++ b/docs/en/rules/Azure.ContainerApp.ExternalAccess.md
@@ -96,6 +96,6 @@ resource containerApp 'Microsoft.App/containerApps@2022-10-01' = {
 
 ## LINKS
 
-- [UNetworking architecture in Azure Container Apps](https://learn.microsoft.com/azure/container-apps/networking)
+- [Networking architecture in Azure Container Apps](https://learn.microsoft.com/azure/container-apps/networking)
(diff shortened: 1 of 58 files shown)


BenjaminEngeset:
@BernieWhite lemme know what you're thinking.

BernieWhite (commented):
@BenjaminEngeset I like the rule. Let me double check a few more details on this to make sure we don't need any exceptions.

BernieWhite (changes requested):
Great work @BenjaminEngeset. I have a few suggestions below.

Also I've been thinking alot about this rule and have the following thoughts:

- External traffic in this case is anything outside the specific container app environment which could be traffic on a private VNET from users or other service consuming the app.
- External traffic is not specifically a bad thing. Most REST services accept external traffic of some kind. The Well-Architected Framework doesn't specifically say you shouldn't use private VNET traffic.
- More importantly the container apps shouldn't be directly exposed to the internet and any traffic should be behind Web Application Firewall.
- The `properties.configuration.ingress.external` = `true` doesn't accurately determine if the container app is exposed to the Internet or only a private VNET. Which is determined by how the environment is deployed with either a public IP load balancer or an internal load balancer.

That said I still like the rule and feel it would be helpful for some customers. So with this in mind, I think we should make this rule opt-in by a configuration option such as `AZURE_CONTAINERAPPS_RESTRICT_INGRESS` = `true` to enable.

What do you think?

src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
```diff
@@ -75,6 +75,26 @@ spec:
...
+kind: Rule
+metadata:
+  name: Azure.ContainerApp.ExternalAccess
+  ref: AZR-000362
+  tags:
+    release: GA
+    ruleSet: 2023_03
+    Azure.WAF/pillar: Security
+  labels:
+    Azure.MCSB.v1/control: NS-2
```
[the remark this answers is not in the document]
BenjaminEngeset: Agreed, it is related to the environment and not the container app. `Configuration Guidance: Disable public network access by deploying an internal-only container apps environment configuration.`

docs/en/rules/Azure.ContainerApp.ExternalAccess.md
```diff
@@ -0,0 +1,101 @@
...
+Ensure inbound communication for Container Apps is limited to callers within the Container Apps environment.
+
+## DESCRIPTION
+
+Container apps allows you to expose your container app to the public web, to your VNET, or to other container apps within your environment by enabling ingress.
+
+Container apps by default will automatically by default have visibility within app environment only if ingress is enabled.
+This secure by default behaviour can be overriden by enabling visibility from internet or VNET, depending on app environment endpoint configured.
+
+Disable external network access to container apps by enforcing internal-only ingress. This will ensure inbound communication for container apps is limited to callers within the container apps environment.
```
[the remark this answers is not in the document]
BernieWhite: My thinking was from if someone deploys it via the Portal (not infrastructure as code) it seems to default to external ingress.

BernieWhite:
> > Great work @BenjaminEngeset. I have a few suggestions below.
> > Also I've been thinking alot about this rule and have the following thoughts:
> > 
> > * External traffic in this case is anything outside the specific container app environment which could be traffic on a private VNET from users or other service consuming the app.
> > * External traffic is not specifically a bad thing. Most REST services accept external traffic of some kind. The Well-Architected Framework doesn't specifically say you shouldn't use private VNET traffic.
> > * More importantly the container apps shouldn't be directly exposed to the internet and any traffic should be behind Web Application Firewall.
> > * The `properties.configuration.ingress.external` = `true` doesn't accurately determine if the container app is exposed to the Internet or only a private VNET. Which is determined by how the environment is deployed with either a public IP load balancer or an internal load balancer.
> > 
> > That said I still like the rule and feel it would be helpful for some customers. So with this in mind, I think we should make this rule opt-in by a configuration option such as `AZURE_CONTAINERAPPS_RESTRICT_INGRESS` = `true` to enable.
> > What do you think?
> 
> @BernieWhite Great feedback, thanks. All points makes sense. Where do we have documentation for implementing configuration option? Possible in YAML?

There isn't an exact example, but use this as a reference to populate a `where` precondition sub-selector: 

- https://microsoft.github.io/PSRule/v2/expressions/functions/#examples
- https://microsoft.github.io/PSRule/v2/expressions/sub-selectors/#pre-conditions

As this is still currently experimental, if you run into any bugs let me know.

BenjaminEngeset:
@BernieWhite check and leets adjust if needed.

src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
```diff
@@ -98,6 +98,25 @@ spec:
...
+  name: Azure.ContainerApp.ExternalIngress
+  ref: AZR-000362
+  tags:
+    release: GA
+    ruleSet: 2023_03
+    Azure.WAF/pillar: Security
+spec:
+  with:
+  - Azure.ContainerApp.ConfigurationExternalIngress
+  - Azure.ContainerApp.WithIngress
```
BernieWhite, line 115: Each entry in `with` is OR instead of AND which is required for this to work correctly. If you use `where` instead that will AND. Also were unlikely to need to reuse the selector for other rules.

```suggestion
  with:
  - Azure.ContainerApp.WithIngress
  where:
    value:
      $:
        configuration: AZURE_CONTAINERAPPS_RESTRICT_INGRESS
    equals: true
```

docs/en/rules/Azure.ContainerApp.ExternalIngress.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+    configuration: {
+      ingress: {
+        external: false
+      }
+    }
+  }
+}
+```
+
```
BernieWhite, line 96: Let's add a note to flag that this rule doesn't execute by default.

```suggestion

## NOTES

This rule is skipped by default because there are common cases where external ingress is required.
If you don't need external ingress, enable this rule by:

- Setting the `AZURE_CONTAINERAPPS_RESTRICT_INGRESS` configuration option to `true`.
```

tests/PSRule.Rules.Azure.Tests/Azure.ContainerApp.Tests.ps1
```diff
@@ -85,6 +85,22 @@ Describe 'Azure.ContainerApp' -Tag 'ContainerApp' {
             $ruleResult | Should -Not -BeNullOrEmpty;
             $ruleResult.Length | Should -Be 1;
             $ruleResult.TargetName | Should -BeIn 'capp-env-C';
+
+        It 'Azure.ContainerApp.ExternalIngress' {
+            $filteredResult = $result | Where-Object { $_.RuleName -eq 'Azure.ContainerApp.ExternalIngress' };
```
BernieWhite, line 90: On line 37 you will need to add:

```powershell
$result = Invoke-PSRule @invokeParams -InputPath $dataPath -Option @{
    'Configuration.AZURE_CONTAINERAPPS_RESTRICT_INGRESS' = $True
}
```

Otherwise this rule will not execute and you should get a test failure.

BernieWhite (changes requested):
Great updates @BenjaminEngeset. Just a few more updates and we should be good.

BenjaminEngeset:
> Great updates @BenjaminEngeset. Just a few more updates and we should be good.

@BernieWhite Thanks for all the feedback and guidance on this one. Check latest updates :)

BenjaminEngeset:
@BernieWhite when you have the time 👍

BernieWhite (approved):
Thanks @BenjaminEngeset. Looks good to me. All good to merge.

PR #2108, opened 2023-03-24
Added Azure.ContainerApp.PersistantStorage

### LINKED ISSUE #2101: Container Apps should use volume mounts for to ensure availability of persistent storage capacity
### Existing rule

_No response_

### Suggested rule

Container Apps should use volume mounts for to ensure availability of persistent storage capacity.

Defaults to using its own container file system, which has no capacity guarantees. The available storage depends on the amount of disk space available in the container.

I guess that `Reliability` is better suited for this than `Operational Excellence` pillar, but open for feedback.

### Pillar

Reliability

### Additional context

- [Use storage mounts in Azure Container Apps](https://learn.microsoft.com/azure/container-apps/storage-mounts)
- [Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.app/containerapps##volume)

BenjaminEngeset:
## PR Summary

Fixes #2101 

Added Azure.ContainerApp.PersistantStorage rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
+++ b/src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
@@ -75,6 +75,28 @@ spec:
       - name: '.'
         match: '^([a-z][0-9a-z-]{0,30}[0-9a-z])$'
 
+---
+# Synopsis: Ensure the use of volume mounts for Container Apps to ensure availability of persistent storage capacity.
+apiVersion: github.com/microsoft/PSRule/v1
+kind: Rule
+metadata:
+  name: Azure.ContainerApp.PersistantStorage
+  ref: AZR-000364
+  tags:
+    release: GA
+    ruleSet: 2023_03
+    Azure.WAF/pillar: Reliability
+spec:
+  level: Warning
+  type:
+  - Microsoft.App/containerApps
+  condition:
+    allOf:
+      - field: properties.template.containers.volumeMounts
+        greater: 0
+      - field: properties.template.volumes
+        greater: 0
+
 #endregion Rules
 
 #region Selectors
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -30,6 +30,8 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
       [#2094](https://github.com/Azure/PSRule.Rules.Azure/issues/2094)
     - Check that managed identity for container apps are configured by @BenjaminEngeset.
       [#2096](https://github.com/Azure/PSRule.Rules.Azure/issues/2096)
+    - Check that volume mounts for container apps are configured by @BenjaminEngeset.
+      [#2101](https://github.com/Azure/PSRule.Rules.Azure/issues/2101)
 
 What's changed since v1.25.0:
 
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -24,6 +24,11 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
 
 ## Unreleased
 
+- New rules:
+  - Container App:
+    - Check that volume mounts for container apps are configured by @BenjaminEngeset.
+      [#2101](https://github.com/Azure/PSRule.Rules.Azure/issues/2101)
+
 ## v1.26.0-B0011 (pre-release)
 
 What's changed since v1.25.0:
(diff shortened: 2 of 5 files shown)


BenjaminEngeset:
@BernieWhite let me know.

BenjaminEngeset:
Just discovered that more than 35 characters for the rule name causing trouble, we could change it to `Azure.ContainerApp.Persistant` or something else?

BernieWhite:
> Just discovered that more than 35 characters for the rule name causing trouble, we could change it to `Azure.ContainerApp.Persistant` or something else?

I think just `Azure.ContainerApp.Storage` may be clearer.

BenjaminEngeset:
> > Just discovered that more than 35 characters for the rule name causing trouble, we could change it to `Azure.ContainerApp.Persistant` or something else?
> 
> I think just `Azure.ContainerApp.Storage` may be clearer.

changed to that :)

BenjaminEngeset:
> Thanks @BenjaminEngeset.
> 
> We are unable to detect if an application needs storage or not, as this is more how the application is coded then how the infrastructure is configured. It is a recommended practise to design stateless applications so normally storage would not be required, but currently this rule requires storage.
> 
> I think we should:
> 
> * Ignore container apps that have no `volumes` or `"volumes": []`.
> * Focus on the cases that have defined volumes (indicating they require storage). Any volumes should be using `AzureFile`.
> 
> Container Apps should be using storage on Azure Files because multiple replicas should have access to the same data so that state if required is not tied to a single instance (i.e. the state is externalized into Azure Files).
> 
> What do you think?

Great feedback and it all makes sense. This recommandation and rule design is definitely preferred over the Azure Policy built-in one. I'll refactor and update you once I think it's ready for review.

BenjaminEngeset:
@BernieWhite Tweaked as discussed, what do you think?

docs/en/rules/Azure.ContainerApp.Storage.md
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+```
+
+### Configure with Bicep
+
+To deploy Container Apps that pass this rule:
+
+- Configure the `properties.template.mounts` array to define a volume or several volumes.
+- For each container in the template that you want to mount storage, define a volume mount in the `properties.template.containers.volumeMounts` array.
```
BernieWhite, line 112:
```suggestion
- For each container in the template that you want to mount storage, define a volume mount in the `properties.template.containers.volumeMounts` array.
- For each volume mount use the `storageType` of `AzureFile`.
```

docs/en/rules/Azure.ContainerApp.Storage.md
```diff
@@ -0,0 +1,164 @@
...
+Consider using Azure File volume mounts to persistent storage across containers and replicas.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To deploy Container Apps that pass this rule:
+
+- Configure the `properties.template.mounts` array to define a volume or several volumes.
+- For each container in the template that you want to mount storage, define a volume mount in the `properties.template.containers.volumeMounts` array.
```
BernieWhite, line 57:
```suggestion
- For each container in the template that you want to mount storage, define a volume mount in the `properties.template.containers.volumeMounts` array.
- For each volume mount use the `storageType` of `AzureFile`.
```

src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
```diff
@@ -98,6 +98,30 @@ spec:
...
+kind: Rule
+metadata:
+  name: Azure.ContainerApp.Storage
+  ref: AZR-000364
+  tags:
+    release: GA
+    ruleSet: 2023_03
+    Azure.WAF/pillar: Reliability
+spec:
+  level: Warning
```
BernieWhite, line 113: I think we can allow this to default to `Error` now that we are only targeting confirmed storage requirements.

What do you think?

BernieWhite (changes requested):
Hi @BenjaminEngeset. A few minor suggestions. If you have any questions let me know.

src/PSRule.Rules.Azure/rules/Azure.ContainerApp.Rule.yaml
BenjaminEngeset: Agreed, specially given how the rule works now.

BenjaminEngeset:
@BernieWhite I've done some small adjustments as I found some typos as well.

I've also added an notes section that informs about something important in order to make this work, which is not needed for ephemeral config.

Will you check?

PR #2218, opened 2023-05-16
Added Azure.Cosmos.DefenderCloud

### LINKED ISSUE #2203: Cosmos DBs should use Microsoft Defender for Azure Cosmos DB
### Existing rule

_No response_

### Suggested rule

Cosmos DBs should use Microsoft Defender for Azure Cosmos DB which is a feature of Microsoft Defender for Cloud.

 Microsoft Defender for Azure Cosmos DB detects potential SQL injections, known bad actors based on Microsoft Threat Intelligence, suspicious access patterns, and potential exploitation of databases through compromised identities, or malicious insiders.

This can be enabled for an individual account by deploying an `Microsoft.Security/advancedThreatProtectionSettings` sub-resource under the parent account.

Currently only Azure Cosmos DB for NoSQL is supported, so we have to make sure the rule only runs against these accounts.

When enabling the Defender pricing plan for Cosmos DB this is automatically enabled for all Azure Cosmos DB for NoSQL accounts within the subscription.

Although the recommended approach is by enabling the pricing plan this rule covers scenarios with assurance (if pricing plan already enabled, nothing is changed), if not it should be enabled for this resource within Defender for Cloud.

So both the upcoming `Azure.Defender.CosmosDbs` and this rule will work nicely together covering both scenarios.

### Pillar

Security

### Additional context

https://learn.microsoft.com/azure/defender-for-cloud/concept-defender-for-cosmos
https://learn.microsoft.com/azure/defender-for-cloud/defender-for-databases-enable-cosmos-protections
https://learn.microsoft.com/azure/templates/microsoft.security/advancedthreatprotectionsettings

BenjaminEngeset:
## PR Summary

Fixes #2203

Added Azure.Cosmos.DefenderCloud rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -323,7 +323,7 @@ What's changed since v1.24.2:
     [#1935](https://github.com/Azure/PSRule.Rules.Azure/issues/1935)
     - Configure this option to increase the visibility of the version of the Bicep CLI used by PSRule for Azure.
     - Set `AZURE_BICEP_CHECK_TOOL` to `true` to check the Bicep CLI.
-    - Set `AZURE_BICEP_MINIMUM_VERSION` to  the minimum version.
+    - Set `AZURE_BICEP_MINIMUM_VERSION` to configure the minimum version.
     - If the Bicep CLI is not installed or the version is less than the minimum version an error will be reported.
     - By default, the minimum Bicep version defaults to `0.4.451`.
   - Added support for Bicep custom types by @BernieWhite.
@@ -400,7 +400,7 @@ What's changed since pre-release v1.25.0-B0013:
 - General improvements:
   - Added support for configuring a minimum version of Bicep by @BernieWhite.
     [#1935](https://github.com/Azure/PSRule.Rules.Azure/issues/1935)
-    -  this option to increase the visibility of the version of the Bicep CLI used by PSRule for Azure.
+    - Configure this option to increase the visibility of the version of the Bicep CLI used by PSRule for Azure.
     - Set `AZURE_BICEP_CHECK_TOOL` to `true` to check the Bicep CLI.
     - Set `AZURE_BICEP_MINIMUM_VERSION` to configure the minimum version.
     - If the Bicep CLI is not installed or the version is less than the minimum version an error will be reported.
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -323,7 +323,7 @@ What's changed since v1.24.2:
     [#1935](https://github.com/Azure/PSRule.Rules.Azure/issues/1935)
     - Configure this option to increase the visibility of the version of the Bicep CLI used by PSRule for Azure.
     - Set `AZURE_BICEP_CHECK_TOOL` to `true` to check the Bicep CLI.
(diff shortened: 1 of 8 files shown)


BenjaminEngeset:
@BernieWhite Let me know what you're thinking. I noticed a unknown property called `EnabledApiTypes`. Is this something you are aware of? I cannot find it, neither in a prior API version.

Since we have this in our template, the pre-condition does not work as it should. Should we build in additional logic for this?

https://github.com/Azure/PSRule.Rules.Azure/blob/478982edcf94df1c138a386cb9ae505e147bc89d/tests/PSRule.Rules.Azure.Tests/Resources.Cosmos.json#L27

BernieWhite:
> @BernieWhite Let me know what you're thinking. I noticed an unknown property called `EnabledApiTypes`. Is this something you are aware of? I cannot find it, neither in a prior API version.
> 
> Since we have this in our template, the pre-condition does not work as it should. Should we build in additional logic for this?
> 
> https://github.com/Azure/PSRule.Rules.Azure/blob/478982edcf94df1c138a386cb9ae505e147bc89d/tests/PSRule.Rules.Azure.Tests/Resources.Cosmos.json#L27

I think this is a runtime property. We shouldn't build any logic around it. We should use capabilities to determine what API is used.

BenjaminEngeset:
> > @BernieWhite Let me know what you're thinking. I noticed an unknown property called `EnabledApiTypes`. Is this something you are aware of? I cannot find it, neither in a prior API version.
> > Since we have this in our template, the pre-condition does not work as it should. Should we build in additional logic for this?
> > https://github.com/Azure/PSRule.Rules.Azure/blob/478982edcf94df1c138a386cb9ae505e147bc89d/tests/PSRule.Rules.Azure.Tests/Resources.Cosmos.json#L27
> 
> I think this is a runtime property. We shouldn't build any logic around it. We should use capabilities to determine what API is used.

@BernieWhite Great, agreed, then the rule is ready for review.

BenjaminEngeset:
@BernieWhite Updated the tests now.

BernieWhite (changes requested):
Great work on the rule @BenjaminEngeset.

I'm thinking this should be opt-in by configuration because for the recommended case it is not required, and there is no additional settings that needs to be set.

Let's add a `AZURE_COSMOS_DEFENDER_PER_ACCOUNT` configuration value.

When set to `true` this rule would be evaluated, otherwise skipped.

BenjaminEngeset:
> Great work on the rule @BenjaminEngeset.
> 
> I'm thinking this should be opt-in by configuration because for the recommended case it is not required, and there is no additional settings that needs to be set.
> 
> Let's add a `AZURE_COSMOS_DEFENDER_PER_ACCOUNT` configuration value.
> 
> When set to `true` this rule would be evaluated, otherwise skipped.

Agreed, that makes the most sense. I'll adjust and ping you after.

BenjaminEngeset:
@BernieWhite I've updated it now and added test case for enablement. Let me know

BernieWhite (approved):
Thanks for the updates @BenjaminEngeset. Looks great. All good to merge.

Issue #2206, opened 2023-05-12
Microsoft Defender for Storage pricing plans should have malware scanning configured


BenjaminEngeset:
### Suggested rule

Microsoft Defender for Storage introduces the malware scanning feature that can be configured within a pricing plan.

Malware scanning in Defender for Storage helps protect storage accounts from malicious content by performing a full malware scan on uploaded content in near real time, using Microsoft Defenders Antivirus capabilities. It's designed to help fulfill security and compliance requirements to handle untrusted content.

An Event Grid System Topic resource is created in the same resource group of the storage account, used by the malware scanning service to listen on blob upload triggers.

Azure Blob Storage is only supported currently. Azure DNS Zone is not supported for malware scanning currently.

Currently this is a preview feature.

This can be enabled by configuring an extension under the subscription pricing plan or by a adding a resource storage account pricing plan to a standalone storage account resource.

Example for subscription pricing plan: 

```
resource StorageAccounts 'Microsoft.Security/pricings@2023-01-01' = {
  name: 'StorageAccounts'
  properties: {
    pricingTier: 'Standard'
    subPlan: 'DefenderForStorageV2'
    extensions: [
      {
        name: 'OnUploadMalwareScanning'
        isEnabled: 'True'
        additionalExtensionProperties: {
          CapGBPerMonthPerStorageAccount: '5000'
        }
      }
    ]
  }
}
```
Example for a specific storage account on a resource pricing plan:

```
resource storageAccount 'Microsoft.Storage/storageAccounts@2021-04-01' ...

resource defenderForStorageSettings 'Microsoft.Security/DefenderForStorageSettings@2022-12-01-preview' = {
  name: 'current'
  scope: storageAccount
  properties: {
    isEnabled: true
    malwareScanning: {
      onUpload: {
        isEnabled: true
        capGBPerMonth: 5000
      }
    }
    overrideSubscriptionLevelSettings: true
  }
}
```

### Pillar

Security

### Additional context

https://learn.microsoft.com/azure/defender-for-cloud/defender-for-storage-malware-scan
https://learn.microsoft.com/azure/storage/common/azure-defender-storage-configure

BenjaminEngeset:
@BernieWhite Two rules? As the plan can be enabled at supscription or resource level. Just have to be clear in docs to not configure any conflicts unless it's done on purpose.

```
[...]

Issue #2207, opened 2023-05-12
Microsoft Defender for Storage pricing plans should have sensitive data threat detection configured


BenjaminEngeset:
### Suggested rule

Microsoft Defender for Storage introduces the sensitive data threat detection feature that can be configured within a pricing plan.

Sensitive data threat detection is powered by the sensitive data discovery engine, an agentless engine that uses a smart sampling method to find resources with sensitive data.

Currently this is a preview feature.

This can be enabled by configuring an extension under the subscription pricing plan or by a adding a standalone storage account to the pricing plan.

Example for subscription pricing plan: 

```
resource StorageAccounts 'Microsoft.Security/pricings@2023-01-01' = {
  name: 'StorageAccounts'
  properties: {
    pricingTier: 'Standard'
    subPlan: 'DefenderForStorageV2'
    extensions: [
      {
        name: 'SensitiveDataDiscovery'
        isEnabled: 'True'
      }
    ]
  }
}
```
Example for a specific storage account to the pricing plan:

```
resource storageAccount 'Microsoft.Storage/storageAccounts@2021-04-01' ...

resource defenderForStorageSettings 'Microsoft.Security/DefenderForStorageSettings@2022-12-01-preview' = {
  name: 'current'
  scope: storageAccount
  properties: {
    sensitiveDataDiscovery: {
      isEnabled: true
    }
    overrideSubscriptionLevelSettings: true
  }
}
```

### Pillar

Security

### Additional context

https://learn.microsoft.com/azure/defender-for-cloud/defender-for-storage-malware-scan
https://learn.microsoft.com/azure/storage/common/azure-defender-storage-configure

### Pillar

Security

### Additional context

https://learn.microsoft.com/azure/defender-for-cloud/defender-for-storage-data-sensitivity
https://learn.microsoft.com/azure/storage/common/azure-defender-storage-configure

BernieWhite:
Same as #2206

This should be two rules.

The recommendation is to apply it at the subscription level, so the main rule should make sure it is turned on with the `Microsoft.Security/pricings` resource.

Then at the storage account level we should just check that that sensitive data discovery is not turned off via `Microsoft.Security/DefenderForStorageSettings`.

PR #2224, opened 2023-05-19
Added Azure.Storage.DefenderCloud.MalwareScan

BenjaminEngeset:
## PR Summary

Part of #2206

Added Azure.Storage.DefenderCloud.MalwareScan rule.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.Storage.DefenderCloud.MalwareScan.md
+++ b/docs/en/rules/Azure.Storage.DefenderCloud.MalwareScan.md
@@ -37,7 +37,8 @@ It is also worth to mention that the resouce level enablement can be useful when
 
 ## RECOMMENDATION
 
-Consider using Malware Scanning in Microsoft Defender for Storage.
+Consider enabling Malware Scanning using Microsoft Defender for Storage on the Storage Account.
+Alternatively, enable Malware Scanning for all Storage Accounts within a subscription.
 
 ## EXAMPLES
 
--- a/docs/en/rules/Azure.Storage.DefenderCloud.MalwareScan.md
+++ b/docs/en/rules/Azure.Storage.DefenderCloud.MalwareScan.md
@@ -1,0 +1,121 @@
+---
+severity: Critical
+pillar: Security
+category: Data protection
+resource: Storage Account
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.Storage.DefenderCloud.MalwareScan/
+---
+
+# Malware Scanning
+
+## SYNOPSIS
+
+Enable Malware Scanning in Microsoft Defender for Storage.
+
+## DESCRIPTION
+
+Microsoft Defender for Storage provides additional security for storage accounts.
+
+One of the features in the Defender for Storage service is malware scanning that is powered by Microsoft Defender Antivirus.
+
+Content uploaded to cloud storage could be malware. Storage accounts can be a malware entry point into the organization and a malware distribution point. To protect organizations from this threat, content in cloud storage must be scanned for malware before it's accessed.
+
+Malware Scanning in Defender for Storage helps protect storage accounts from malicious content by performing a full malware scan on uploaded content in near real time.
+
+This can be helpful when:
+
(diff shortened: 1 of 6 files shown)


BenjaminEngeset:
@BernieWhite Lemme know and as well:

- Notes section about behaviour, does it look correct to you? I've done some testing and trying to interpret the documentation, but the documentation is a bit unclear about if it really creates it's own dedicated plan no matter what. Please check.
- I see we also have to create a general seperate rule for just Defender on resource level in order for this and #2207 to make the best sense?

src/PSRule.Rules.Azure/rules/Azure.Storage.Rule.ps1
```diff
@@ -88,6 +88,13 @@ Rule 'Azure.Storage.ContainerSoftDelete' -Ref 'AZR-000289' -Type 'Microsoft.Stor
     }
 }
 
+# Synopsis: Enable Malware Scanning in Microsoft Defender for Storage.
+Rule 'Azure.Storage.DefenderCloud.MalwareScan' -Ref 'AZR-000384' -Type 'Microsoft.Storage/storageAccounts' -Tag @{ release = 'Preview'; ruleSet = '2023_06'; 'Azure.WAF/pillar' = 'Security'; } -Labels @{ 'Azure.MCSB.v1/control' = 'DP-2', 'LT-1' } {
+    $malwareConfigured = @(GetSubResources -ResourceType 'Microsoft.Security/DefenderForStorageSettings' |
+        Where-Object { $_.properties.malwareScanning.onUpload.isEnabled -eq $True })
+    $Assert.GreaterOrEqual($malwareConfigured, '.', 1).Reason($LocalizedData.ResStorageMalwareScanning, $PSRule.TargetName)
```
BernieWhite, line 95: We want to find cases when the malware scanning has been disabled. Try the following instead.

```suggestion
    $malwareDisabled = @(GetSubResources -ResourceType 'Microsoft.Security/DefenderForStorageSettings' |
        Where-Object { $_.properties.malwareScanning.onUpload.isEnabled -eq $False })
    $Assert.Count($malwareDisabled, '.', 0).Reason($LocalizedData.ResStorageMalwareScanning, $PSRule.TargetName)
```

Additionally, I think we need to a pre-filter to skip the rule if `properties.publicNetworkAccess` is set to `Disabled` which is not supported as mentioned.

docs/en/rules/Azure.Storage.DefenderCloud.MalwareScan.md
```diff
@@ -0,0 +1,121 @@
...
+
+Malware Scanning in Microsoft Defender for Storage can be enabled at the resource level. However, the general recommandation is to enable it at the subscription level and by doing so ensures all storage accounts in the subscription will be protected, including future ones. Defender for Storage settings on each storage account is inherited by the subscription level settings.
+
+It is also worth to mention that the resouce level enablement can be useful when:
+
+- Override subscription level settings to configure specific storage accounts with custom malware scanning settings that differ from the settings configured at the subscription level.
+
+## RECOMMENDATION
+
+Consider using Malware Scanning in Microsoft Defender for Storage.
```
BernieWhite, line 40:
```suggestion
Consider enabling Malware Scanning using Microsoft Defender for Storage on the Storage Account.
Alternatively, enable Malware Scanning for all Storage Accounts within a subscription.
```

BernieWhite (changes requested):
Thanks @BenjaminEngeset. A few suggested changes. Separate rule is good.

Let me know if that doesn't answer your question.

BenjaminEngeset:
Hi @BernieWhite I have done some adjustments and I'd like you to check them. As the rule is working now, only storage accounts with sub-resource set to disabled fails and else passes, but I guess that is what you had in mind, and rather use the subscription rules as the mains, right?

BernieWhite:
Correct only storage accounts explicitly disabling the malware scan should fail.

BernieWhite (approved):
Thanks @BenjaminEngeset . All good to merge.

PR #2237, opened 2023-05-23
Fixed false positive on managed identity #2235

### LINKED ISSUE #2235: [Bug] Azure.APIM.ManagedIdentity when SystemAssigned and UserAssigned Identities are enabled
### Existing rule

Azure.APIM.ManagedIdentity

### Suggested rule

Azure.APIM.ManagedIdentity rule fails with the following error when both UserAssigned and SystemAssigned Managed identities are enabled on the resource.

Recommend:
  Consider configuring a managed identity for each API Management instance. Also
  consider using managed identities to authenticate to related Azure services.

  Reason:
  - Path Identity.Type: The value 'SystemAssigned,UserAssigned' was not in 'System.String[]'.

### Pillar

Security

### Additional context

_No response_

BernieWhite:
## PR Summary

- Fixed false positive on managed identity without space.

Fixes #2235 

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue

### DIFF
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -24,6 +24,12 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
 
 ## Unreleased
 
+What's changed since pre-release v1.27.0-B0091:
+
+- Bug fixes:
+  - Fixed false positive on managed identity without space by @BernieWhite.
+    [#2235](https://github.com/Azure/PSRule.Rules.Azure/issues/2235)
+
 ## v1.27.0-B0091 (pre-release)
 
 What's changed since pre-release v1.27.0-B0050:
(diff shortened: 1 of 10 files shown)


BenjaminEngeset:
@BernieWhite Just out of curiosity, is this because PSRule treats a comma seperated list with spaces between as an array and without the spaces it is the explicit string value?

BernieWhite:
@BenjaminEngeset The `in` assertion/ expression is any of the literal values (case insensitive). It doesn't break by commas.

BenjaminEngeset:
> @BenjaminEngeset The `in` assertion/ expression is any of the literal values (case insensitive). It doesn't break by commas.

Thanks for the explanation @BernieWhite!

PR #2756, opened 2024-03-15
Update Azure.LB.Rule.ps1 for PIP Zones Condition

jtracey93:
## PR Summary

This pull request includes a change to the `Azure.LB.AvailabilityZone` rule in the `Azure.LB.Rule.ps1` file. The change removes the condition that checks if the `zones` property of the `ipConfig` object is null or empty. This means that the rule will now only assert if the `zones` property is a set of '1', '2', '3'.

This is what needs to be true for the LB to be truly zone redundant

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [ ] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/tests/PSRule.Rules.Azure.Tests/Resources.VirtualNetwork.json
+++ b/tests/PSRule.Rules.Azure.Tests/Resources.VirtualNetwork.json
@@ -1736,7 +1736,10 @@
               }
             ],
             "privateIPAddressVersion": "IPv4"
-          }
+          },
+          "zones": [
+            "3"
+          ]
         }
       ],
       "backendAddressPools": [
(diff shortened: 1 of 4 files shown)


tests/PSRule.Rules.Azure.Tests/Resources.VirtualNetwork.json
```diff
@@ -1731,10 +1736,7 @@
               }
             ],
             "privateIPAddressVersion": "IPv4"
-          },
-          "zones": [
-            "3"
-          ]
+          }
```
BernieWhite, line 1739: I think we should still check and fail for the case where a zonal LB is configured.


```suggestion
          },
          "zones": [
            "3"
          ]
```

BernieWhite (changes requested):
@jtracey93 Great work, and good spot.

I've added checks for the case when AZ is not supported in the region.

Otherwise double check my minor comments, and let me know if you're good with my changes.

jtracey93:
all good @BernieWhite

BernieWhite (approved):
@jtracey93 Awesome.

PR #2785, opened 2024-03-29
Updated links to learn.microsoft.com (from docs.microsoft.com)

lukemurraynz:
Still a few references to docs.microsoft.com.

Even though this works, it is a redirect and could potentially break in the future.

Cleaning up links to point to learn.microsoft.com directly.


BernieWhite (approved):
Thanks @lukemurraynz. All good to merge.

PR #2872, opened 2024-05-19
feat(new): Added Azure.EventHub.Firewall

### LINKED ISSUE #2701: [RULE] Event Hub Firewall
### Existing rule

_No response_

### Suggested rule

Access to Event Hub endpoints should be restricted to only allowed sources.

- Ideally use private endpoints. OR
- Configure the firewall in default `deny` mode.

### Pillar

Security

### Additional context

- https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-ip-filtering
- https://learn.microsoft.com/en-us/azure/event-hubs/private-link-service
- https://learn.microsoft.com/en-us/azure/event-hubs/policy-reference
- https://learn.microsoft.com/en-us/security/benchmark/azure/baselines/event-hubs-security-baseline?toc=%2Fazure%2Fevent-hubs%2FTOC.json#disable-public-network-access

BenjaminEngeset:
## PR Summary

Fixes #2701 

Added Azure.EventHub.Firewall.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


BenjaminEngeset:
Hi @BernieWhite. I want some feedback from you on this one, what do you think? The `AnyOf` method used can give customers some output that can be confusing, but after reading the rule documentation I suppose it will make sense?

I can also go for another design with an `if else` statement using `$PSRule.TargetType`, but will be some "duplicated" code by doing so for each block.

BernieWhite:
@BenjaminEngeset Nice work so far. This one does have one additional complexity that you may not have encountered with previous rules, that I hadn't thought about in the initial issue.

There is an order of precedence on which will take effect for the final resource.

ARM take the approach of deploy the parent resource first `Microsoft.EventHub/namespaces` then the sub-resource `Microsoft.EventHub/namespaces/networkRuleSets`.

The impact of this is: **If the sub-resource is set it will override the configuration on the parent resource, because it is deployed last and modifies the state of the parent.**

Use `Azure.SQL.AAD` rule as a guide for the implementation: https://github.com/benjaminengeset/psrule.rules.azure/blob/daf85d5b183913461c6ef4d575e8fe93aae727a6/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1#L58-L80

i.e.

- If the resource type is `Microsoft.EventHub/namespaces/networkRuleSets` check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
- If the resource type is `Microsoft.EventHub/namespaces` check for sub-resources an loop over any found performing a check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
- Fall back to check of the parent resource properties `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
-

docs/en/rules/Azure.EventHub.Firewall.md
```diff
@@ -0,0 +1,99 @@
...
+
+Consider configuring Event Hub namespaces to disable public network access, using private endpoints to control connectivity or configure the firewall to only allow traffic from from specific addresses.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To deploy Event Hub namespaces that pass this rule:
+
+- Set the `properties.publicNetworkAccess` property to `Disabled` or configure the sub-resource `Microsoft.EventHub/namespaces/networkRuleSets` with the `properties.publicNetworkAccess` property to `Disabled` or/and the `properties.defaultAction` property to `Deny`.
```
BernieWhite, line 31:
```suggestion
- Set the `properties.publicNetworkAccess` property to `Disabled` to require private endpoints. OR
- Alternatively, you can configure the `Microsoft.EventHub/namespaces/networkRuleSets` sub-resource by:
  - Setting the `properties.publicNetworkAccess` property to `Disabled` to require private endpoints. OR
  - Setting the `properties.defaultAction` property to `Deny` to restrict network access to the service by default.
```

docs/en/rules/Azure.EventHub.Firewall.md
```diff
@@ -0,0 +1,99 @@
...
+    "zoneRedundant": true
+  }
+}
+```
+
+### Configure with Bicep
+
+To deploy Event Hub namespaces that pass this rule:
+
+- Set the `properties.publicNetworkAccess` property to `Disabled` or configure the sub-resource `Microsoft.EventHub/namespaces/networkRuleSets` with the `properties.publicNetworkAccess` property to `Disabled` or/and the `properties.defaultAction` property to `Deny`.
```
BernieWhite, line 60:
```suggestion
- Set the `properties.publicNetworkAccess` property to `Disabled` to require private endpoints. OR
- Alternatively, you can configure the `Microsoft.EventHub/namespaces/networkRuleSets` sub-resource by:
  - Setting the `properties.publicNetworkAccess` property to `Disabled` to require private endpoints. OR
  - Setting the `properties.defaultAction` property to `Deny` to restrict network access to the service by default.
```

docs/en/rules/Azure.EventHub.Firewall.md
```diff
@@ -0,0 +1,99 @@
...
+
+## DESCRIPTION
+
+By default, Event Hub namespaces are accessible from public internet.
+
+With the firewall feature, it is possible to either fully disabling public network access by ensuring that the namespace endpoints isn't exposed on the public internet or configure rules to only accept traffic from specific addresses.
+
+## RECOMMENDATION
+
+Consider configuring Event Hub namespaces to disable public network access, using private endpoints to control connectivity or configure the firewall to only allow traffic from from specific addresses.
```
BernieWhite, line 23:
```suggestion
Consider restricting network access to the Event Hub namespace by requiring private endpoints or by limiting access to permitted client addresses with the service firewall.
```

BernieWhite (commented):
Some additional suggestions for docs.

BenjaminEngeset:
> https://github.com/benjaminengeset/psrule.rules.azure/blob/daf85d5b183913461c6ef4d575e8fe93aae727a6/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1#L58-L80

Great feedback. I knew this initially, but wasn't sure if we should have such logic in the rule body. Well now I know and I will adjust accordingly. Thanks for the great feedback.

BenjaminEngeset:
> @BenjaminEngeset Nice work so far. This one does have one additional complexity that you may not have encountered with previous rules, that I hadn't thought about in the initial issue.
> 
> There is an order of precedence on which will take effect for the final resource.
> 
> ARM take the approach of deploy the parent resource first `Microsoft.EventHub/namespaces` then the sub-resource `Microsoft.EventHub/namespaces/networkRuleSets`.
> 
> The impact of this is: **If the sub-resource is set it will override the configuration on the parent resource, because it is deployed last and modifies the state of the parent.**
> 
> Use `Azure.SQL.AAD` rule as a guide for the implementation: https://github.com/benjaminengeset/psrule.rules.azure/blob/daf85d5b183913461c6ef4d575e8fe93aae727a6/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1#L58-L80
> 
> i.e.
> 
> * If the resource type is `Microsoft.EventHub/namespaces/networkRuleSets` check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
> * If the resource type is `Microsoft.EventHub/namespaces` check for sub-resources an loop over any found performing a check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
> * Fall back to check of the parent resource properties `properties.publicNetworkAccess`. Using `AnyOf` is fine.

Makes sense. I was aware of this, but didn't expect us to take that into consideration, but now that we are aligned I'll adjust accordingly. Great feedback. The `$Assert.AnyOf()` method doesn't work so well for me. 

image
image

Writing it with `AnyOf{}` instead works. What is this, a global function?

image

Found it here:

https://github.com/Azure/PSRule.Rules.Azure/blob/b65fb70311dba052f862ebda9db777939279246e/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1#L131

BernieWhite:
> > @BenjaminEngeset Nice work so far. This one does have one additional complexity that you may not have encountered with previous rules, that I hadn't thought about in the initial issue.
> > There is an order of precedence on which will take effect for the final resource.
> > ARM take the approach of deploy the parent resource first `Microsoft.EventHub/namespaces` then the sub-resource `Microsoft.EventHub/namespaces/networkRuleSets`.
> > The impact of this is: **If the sub-resource is set it will override the configuration on the parent resource, because it is deployed last and modifies the state of the parent.**
> > Use `Azure.SQL.AAD` rule as a guide for the implementation: https://github.com/benjaminengeset/psrule.rules.azure/blob/daf85d5b183913461c6ef4d575e8fe93aae727a6/src/PSRule.Rules.Azure/rules/Azure.SQL.Rule.ps1#L58-L80
> > i.e.
> > 
> > * If the resource type is `Microsoft.EventHub/namespaces/networkRuleSets` check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
> > * If the resource type is `Microsoft.EventHub/namespaces` check for sub-resources an loop over any found performing a check `properties.publicNetworkAccess` or `properties.defaultAction`. Using `AnyOf` is fine.
> > * Fall back to check of the parent resource properties `properties.publicNetworkAccess`. Using `AnyOf` is fine.
> 
> Makes sense. I was aware of this, but didn't expect us to take that into consideration, but now that we are aligned I'll adjust accordingly. Great feedback. The `$Assert.AnyOf()` method doesn't work so well for me.
> 
> image image
> Writing it with `AnyOf{}` instead works. What is this, a global function?
> 
> image
> Found it here:
> 
> https://github.com/Azure/PSRule.Rules.Azure/blob/b65fb70311dba052f862ebda9db777939279246e/src/PSRule.Rules.Azure/rules/Azure.Template.Rule.ps1#L131

`$Assert.AnyOf` would require `,` between because it is an array of results. See: https://microsoft.github.io/PSRule/v2/concepts/PSRule/en-US/about_PSRule_Assert/#aggregating-assertion-methods

Or use the `{ }` if that works better for you. Either is fine.

BernieWhite:
> Should be ready for another review @BernieWhite. I don't undestand why `$ruleresult[2].Reason` and up to `$ruleresult[4].Reason` is empty. Any idea why the property has no value on these objects?

I did a bit of investigating and this seems to be a bug with `$Assert.AnyOf`, it is currently dropping any aggregated reasons. Using `AnyOf` produces the correct result.

BernieWhite (approved):

PR #2905, opened 2024-06-01
feat: Updated Azure.MySQL.DefenderCloud

### LINKED ISSUE #2904: [RULE] Add support for Defender for Cloud for Azure Database for MySQL for the flexible deployment model
### Existing rule

Azure.MySQL.DefenderCloud

### Suggested rule

The flexible deployment model now supports Defender for Cloud for Azure Database for MySQL.

It is done by configuring `Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings`.

This resource resource is read-only.

It can be enabled through either of the following methods:

- Subscription Level Enablement: Enable the `open-source relational databases` resource type under the Microsoft Defender for Cloud Databases plan for the subscription where the server is located.
- Resource Level Enablement: Manually enable Microsoft Defender for Cloud for the specific resource via the Azure portal under the resource blade.

### Pillar

Security

### Additional context

https://azure.microsoft.com/updates/generally-available-defender-for-cloud-supports-azure-database-for-mysql-flexible-server

BenjaminEngeset:
## PR Summary

Fixes #2904

Updated Azure.MySQL.DefenderCloud to also support the flexible deployment model.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -39,6 +39,12 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
   - Virtual Machine Scale Sets:
     - Check that automatic instance repairs are enabled by @BenjaminEngeset.
       [#2895](https://github.com/Azure/PSRule.Rules.Azure/issues/2895)
+- Updated rules:
+  - Azure MySQL:
+    - Updated `Azure.MySQL.DefenderCloud` to support flexible servers by @BenjaminEngeset.
+      [#2788](https://github.com/Azure/PSRule.Rules.Azure/issues/2788)
+      - Added check for Microsoft Defender for Cloud for Azure Database for MySQL for the flexible deployment model.
+      - Bumped rule set to `2024_06`.
 - Engineering:
   - Bump xunit to v2.8.1.
     [#2892](https://github.com/Azure/PSRule.Rules.Azure/pull/2892)
--- a/docs/en/rules/Azure.MySQL.DefenderCloud.md
+++ b/docs/en/rules/Azure.MySQL.DefenderCloud.md
@@ -1,7 +1,8 @@
 ---
+reviewed: 2024-06-01
 severity: Important
 pillar: Security
-category: Security operations
+category: SE:10 Monitoring and threat detection
 resource: Azure Database for MySQL
 online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.MySQL.DefenderCloud/
 ---
@@ -14,7 +15,8 @@ Enable Microsoft Defender for Cloud for Azure Database for MySQL.
 
 ## DESCRIPTION
 
-Defender for Cloud detects anomalous activities indicating unusual and potentially harmful attempts to access or exploit databases.
+Microsoft Defender for Cloud detects anomalous activities that may indicate unusual and potentially harmful attempts to access or exploit your databases. 
+It provides advanced threat protection for your Azure Database for MySQL, helping to safeguard your data and maintain compliance.
 
 ## RECOMMENDATION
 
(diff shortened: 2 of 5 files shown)


BenjaminEngeset:
Hi @BernieWhite. Check out, I've used quite some time to figure out how this works as the resource is read-only for us.

 it is configured only via the portal for the resource within the resource blade `(resource-level enablement)` or via enabling the Defender servers plan (with the open-source relational databases  resource type enabled) for the subscription where the flexible server is located at `(subscription-level enablement)`.

Since this is the behavior, it will only run against exported data for the flexible servers.

`Export-AzRuleData` does not currently support exporting out the `Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings` resource so we have to add support for that, I have tested it out.

BernieWhite:
> Hi @BernieWhite. Check out, I've used quite some time to figure out how this works as the resource is read-only for us.
> 
> it is configured only via the portal for the resource within the resource blade `(resource-level enablement)` or via enabling the Defender servers plan (with the open-source relational databases resource type enabled) for the subscription where the flexible server is located at `(subscription-level enablement)`.
> 
> Since this is the behavior, it will only run against exported data for the flexible servers.
> 
> `Export-AzRuleData` does not currently support exporting out the `Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings` resource so we have to add support for that, I have tested it out.

I'm going see if I can investigate this more internally. The REST API includes PUT operations so I don't think the sub-resource is readonly, it may be a documentation error with the spec.

https://learn.microsoft.com/en-us/rest/api/mysql/flexibleserver/advanced-threat-protection-settings/update-put?view=rest-mysql-flexibleserver-2023-12-30&tabs=HTTP

BenjaminEngeset:
> > Hi @BernieWhite. Check out, I've used quite some time to figure out how this works as the resource is read-only for us.
> > it is configured only via the portal for the resource within the resource blade `(resource-level enablement)` or via enabling the Defender servers plan (with the open-source relational databases resource type enabled) for the subscription where the flexible server is located at `(subscription-level enablement)`.
> > Since this is the behavior, it will only run against exported data for the flexible servers.
> > `Export-AzRuleData` does not currently support exporting out the `Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings` resource so we have to add support for that, I have tested it out.
> 
> I'm going see if I can investigate this more internally. The REST API includes PUT operations so I don't think the sub-resource is readonly, it may be a documentation error with the spec.
> 
> https://learn.microsoft.com/en-us/rest/api/mysql/flexibleserver/advanced-threat-protection-settings/update-put?view=rest-mysql-flexibleserver-2023-12-30&tabs=HTTP

Doesn't look very readonly to me.

The thing is that the Bicep linter yields that it can be only used with the existing keyword.

Let me know what you are able to find out and we'll adjust accordingly.

```
Resource type "Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings@2023-10-01-preview" can only be used with the 'existing' keyword.bicep(BCP245)
```

BernieWhite:
@BenjaminEngeset I've asked the question internally, let's park it for now. Should have a clear path in the next day or two.

BernieWhite:
@BenjaminEngeset there is still some ongoing discussions on this one, hope to have an update soon.

BenjaminEngeset:
What is the current status here @BernieWhite? Still awaiting feedback?

polatengin:
Picking this thread back up, apologies for the long silence here, @BenjaminEngeset.

A lot has changed since this was opened, so before we land it I think it's worth a quick re-check rather than merging on the original assumptions:

- Re-validate the read-only behaviour against the current API version. The original BCP245 was on `Microsoft.DBforMySQL/flexibleServers/advancedThreatProtectionSettings@2023-10-01-preview`. There are newer API versions available now, and the REST PUT operation has always been documented ver good. If the latest stable API version is now deployable via Bicep/ARM, we should support the in-template path as well, not only the export path, otherwise users authoring IaC won't get a result until they export from a live subscription.
- If it is still effectively read-only for IaC, the current export-only approach is the right shape, but we should also fail the parent `Microsoft.DBforMySQL/flexibleServers` (gated on `IsExport`) when the ATP child wasn't exported. Otherwise a flexible server that has never had Defender configured produces no result at all, which is the most security-relevant case. This matches how the single-server path already behaves.
- Rule set version needs bumping before merge. `2024_06` is now ~2 years stale; the next quarterly target is `2026_06`. The changelog entry and the reviewed: frontmatter date should move with it.

If you're happy to rebase and address (3) plus take a look at (1), I can help review again quickly. For (2), happy to take it as a follow-up issue if you'd prefer to keep this PR scoped.

Whatyou think @BernieWhite?

PR #2948, opened 2024-06-18
feat(new): Added Azure.VMSS.ZoneBalance

### LINKED ISSUE #2901: [RULE] Disable Force strictly even balance across zones in Azure Virtual Machine Scale Sets
### Existing rule

_No response_

### Suggested rule

As per [APRL recommendation](https://azure.github.io/Azure-Proactive-Resiliency-Library-v2/azure-resources/Compute/virtualMachineScaleSets/#disable-force-strictly-even-balance-across-zones-to-avoid-scale-in-and-out-fail-attempts), disabling strictly even VM instance distribution across Availability Zones in VMSS improves scalability and flexibility, noting that uneven distribution may better serve application load demands despite the potential trade-off in resilience.

Resource Type: **_microsoft.compute/virtualmachinescalesets_**
Properties should **NOT** have the following values: **_properties.orchestrationMode == "Uniform"_** and **_properties.zoneBalance == true**

### Pillar

Reliability

### Additional context

[VMSS Custom Scale-in Policy in Reliability Hub](https://learn.microsoft.com/en-us/azure/reliability/reliability-virtual-machine-scale-sets?tabs=graph-4%2Cgraph-1%2Cgraph-2%2Cgraph-3%2Cgraph-5%2Cgraph-6%2Cportal#-set-virtual-machine-scale-sets-custom-scale-in-policies-to-default)

This rule will be used to test against the AVM Bicep VMSS module.
@rahalan

BenjaminEngeset:
## PR Summary

Fixes #2901

Added Azure.VMSS.ZoneBalance.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.VMSS.ZoneBalance.md
+++ b/docs/en/rules/Azure.VMSS.ZoneBalance.md
@@ -10,7 +10,7 @@ online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.VMSS.Z
 
 ## SYNOPSIS
 
-Deploy virtual machine scale set instances using the best-effort zone balance in supported regions to ensure an approximately balanced distribution of instances across availability zones.
+Deploy virtual machine scale set instances using the best-effort zone balance in supported regions.
 
 ## DESCRIPTION
 
@@ -22,8 +22,8 @@ The distribution behavior of instances across zones can be configured with two m
 
 - **Best-effort zone balance** - Attempt to scale-in/ out while maintaining balance across zones.
   If that is not possible, allow temporary imbalance so the scaling operation can complete.
-  On subsequent scale-out attempts, the scale set adds VMs to the zones needing more VMs to restore balance.
-  Similarly, on subsequent scale-in attempts, the scale set removes VMs from zones that have more VMs than required to restore balance.
+  - On subsequent scale-out attempts, the scale set adds VMs to the zones needing more VMs to restore balance.
+  - On subsequent scale-in attempts, the scale set removes VMs from zones that have more VMs than required to restore balance.
 - **Strict zone balance** - Only allow the scaling operation to continue if zone balanced can be maintained.
   If that is not possible, the scaling operation will fail.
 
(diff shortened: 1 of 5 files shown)


BenjaminEngeset:

I was unsure if this rule should be triggered when the region does not support availability zones, but if it should, let me know and I'll remove the function.

docs/en/rules/Azure.VMSS.ZoneBalance.md
```diff
@@ -0,0 +1,94 @@
...
+## DESCRIPTION
+
+When deploying virtual machine scale sets across multiple availability zones, the distribution behavior of instances can be configured with two modes:
+
+- Best-effort zone balance
+- Strict zone balance
+
+Best-effort zone balance: This mode attempts to scale in and out while maintaining a balance across zones. If maintaining zone balance is not possible (e.g., if a zone goes down and new VMs can’t be created in that zone), the scale set allows temporary imbalance to continue scaling operations. On subsequent scale-out attempts, the scale set adds VMs to the zones needing more VMs to restore balance. Similarly, on subsequent scale-in attempts, the scale set removes VMs from zones that have more VMs than required to restore balance.
+
+Strict zone balance: This mode ensures that VM instances are strictly balanced across all specified zones, without allowing any temporary imbalance. This can cause the scaling operations to be paused or fail if the zone balance cannot be maintained.
```
BernieWhite, line 24:
```suggestion
Virtual machine scale sets (VMSS) scale-in/ out based on scaling rules your configured.
When scaling-out (adding instances) using availability zones instances are distribution over the configured zones.
However, it may not be possible to create an instance in a zone if there is an active issue affecting that zone.

The distribution behavior of instances across zones can be configured with two modes:

- **Best-effort zone balance** - Attempt to scale-in/ out while maintaining balance across zones.
  If that is not possible, allow temporary imbalance so the scaling operation can complete.
  On subsequent scale-out attempts, the scale set adds VMs to the zones needing more VMs to restore balance.
  Similarly, on subsequent scale-in attempts, the scale set removes VMs from zones that have more VMs than required to restore balance.
- **Strict zone balance** - Only allow the scaling operation to continue if zone balanced can be maintained.
  If that is not possible, the scaling operation will fail.

Key points:

- Both modes attempt to keep balance across zones but have a different approach when balance can't be maintained.
- Scale-out typically occurs to reduce pressure on the already provisioned instances by increasing capacity.
- If scale-out fails, the workload may become unstable due to increasing pressure.
- Balance only applies when two or more zones are configured on the VMSS.
- An outage or disruption in a zone may impact a higher percentage of the workload instances when the workload is not blanced.
```

docs/en/rules/Azure.VMSS.ZoneBalance.md
```diff
@@ -0,0 +1,94 @@
...
+  }
+  zones: [
+    '1'
+    '2'
+    '3'
+  ]
+}
+```
+
+## LINKS
```
BernieWhite, line 90:
```suggestion
<!-- external:avm res/compute/virtual-machine-scale-set zoneBalance -->

## LINKS
```

BernieWhite (changes requested):
Looks great @BenjaminEngeset. A few suggestions for documentation are included below.

BenjaminEngeset:
Thanks for the great feedback @BernieWhite. I have implemented the requested changes now.

BernieWhite (approved):

Issue #2988, opened 2024-07-13
[RULE] Geo-replication for service bus namespaces


BenjaminEngeset:
### Suggested rule

Service bus namespaces supports geo-replication. This feature is currently in preview and has many limitations that will not work for many customers with their current or planned implementation. I think it is still a useful rule to have and we have to adjust it over time when more functionality becomes available.

The geo-replication feature ensures that the metadata and data of a namespace are continuously replicated from a primary region to one or more secondary regions.

The secondaries act as hot stand-by regions and must be promoted to be used.

The following will be replicated:
- Queues, topics, subscriptions, filters.
- Data, which resides in the entities.
- All state changes and property changes executed against the messages within a namespace.
- Namespace configuration.

### Pillar

Reliability

### Additional context

https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-geo-replication

PR #2989, opened 2024-07-13
feat(new): Added Azure.ServiceBus.GeoReplica

BenjaminEngeset:
## PR Summary

Working towards #2988

Added Azure.ServiceBus.GeoReplica.

## PR Checklist

- [ ] Summarized changes
- [ ] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/en/rules/Azure.ServiceBus.GeoReplica.md
+++ b/docs/en/rules/Azure.ServiceBus.GeoReplica.md
@@ -83,7 +83,8 @@ For example:
 
 To deploy namespaces that pass this rule:
 
-- Set `sku.name` to `Premium`.
+- Set the `sku.name` property to `Premium`. This only applies to new namespaces.
+  - For existing namespaces first, migrate to a namespace deployed with a premium SKU.
 - Configure the `properties.geoDataReplication.locations` array with two or more supported region elements.
 
 For example:
--- a/docs/en/rules/Azure.ServiceBus.GeoReplica.md
+++ b/docs/en/rules/Azure.ServiceBus.GeoReplica.md
@@ -127,4 +127,5 @@ This feature is currently in preview.
 - [RE:05 Redundancy](https://learn.microsoft.com/azure/well-architected/reliability/redundancy)
 - [Geo-replication](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-geo-replication)
 - [Configure geo-replication](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-geo-replication#setup)
+- [Migrate existing Azure Service Bus standard namespaces to the premium tier](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-migrate-standard-premium)
 - [Azure deployment reference](https://learn.microsoft.com/azure/templates/microsoft.servicebus/namespaces)
--- a/docs/en/rules/Azure.ServiceBus.GeoReplica.md
+++ b/docs/en/rules/Azure.ServiceBus.GeoReplica.md
@@ -1,0 +1,128 @@
+---
+severity: Important
+pillar: Reliability
+category: RE:05 Redundancy
+resource: Service Bus
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.ServiceBus.GeoReplica/
+---
+
+# Geo-replication
+
+## SYNOPSIS
+
+Enhance resilience to regional outages by replicating namespaces.
+
+## DESCRIPTION
+
+By default, an service bus namespace and its data and metadata is in a single region.
+
+The geo-replication feature ensures that the metadata and data of a namespace are continuously replicated from a primary region to one or more secondary regions.
+
+The following will be replicated:
+
+- Queues, topics, subscriptions, filters.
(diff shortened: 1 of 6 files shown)


BenjaminEngeset:
Hi @BernieWhite. I need some feedback from you on how to construct the rule body on this in the best way possible, also given the updates that will come so we don't have to do major refactor. The documentation states specifically `Currently only a single secondary is supported.`

Do you have any thoughts on how the rule body can look like? I have uploaded the draft design in this PR, but I feel it is to static and if the order is wrong in the array it will not work correctly.

https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-geo-replication#using-bicep-template

BernieWhite:
> Hi @BernieWhite. I need some feedback from you on how to construct the rule body on this in the best way possible, also given the updates that will come so we don't have to do major refactor. The documentation states specifically `Currently only a single secondary is supported.`
> 
> Do you have any thoughts on how the rule body can look like? I have uploaded the draft design in this PR, but I feel it is to static and if the order is wrong in the array it will not work correctly. Maybe two contains methods can be used? If the operand is an array of strings, only one string must contain the specified string.
> 
> https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-geo-replication#using-bicep-template

Thanks @BenjaminEngeset I think the can primarily count how many entries are in `properties.geoDataReplication.locations`. i.e should be 2 or greater. The specific locations or roles and their order don't really matter and could easily change.

BenjaminEngeset:

I noticed one thing that made me unsure about using YAML and that was that once the `AllOf` operator finds a fail, it will return and yield only that. So when the customer thinks that it is only the SKU that is the problem, next time it will yield the geo-replication configuration. I want both failures to be yielded. What do you think, is there anything we can do or should we move over to PowerShell?

BernieWhite:
> Ready for review @BernieWhite.
> 
> I noticed one thing that made me unsure about using YAML and that was that once the `AllOf` operator finds a fail, it will return and yield only that. So when the customer thinks that it is only the SKU that is the problem, next time it will yield the geo-replication configuration. I want both failures to be yielded. What do you think, is there anything we can do or should we move over to PowerShell?

That's fair. We should just improve that behaviour in PSRule.

docs/en/rules/Azure.ServiceBus.GeoReplica.md
```diff
@@ -0,0 +1,128 @@
...
+
+Consider replicating service bus namespaces to improve resiliency to region outages.
+
+## EXAMPLES
+
+### Configure with Azure template
+
+To deploy namespaces that pass this rule:
+
+- Set `sku.name` to `Premium`.
```
BernieWhite, line 47: It's not possible to easily set the SKU to `Premium` if it's not already premium. There is a migration required to a new Premium namespace from an existing Standard tier.

```suggestion
- Set the `sku.name` property to `Premium`. This only applies to new namespaces.
  - For existing namespaces first, migrate to a namespace deployed with a premium SKU.
```

docs/en/rules/Azure.ServiceBus.GeoReplica.md
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+}
+```
+
+### Configure with Bicep
+
+To deploy namespaces that pass this rule:
+
+- Set `sku.name` to `Premium`.
```
BernieWhite, line 85: It's not possible to easily set the SKU to `Premium` if it's not already premium. There is a migration required to a new Premium namespace from an existing Standard tier.

```suggestion
- Set the `sku.name` property to `Premium`. This only applies to new namespaces.
  - For existing namespaces first, migrate to a namespace deployed with a premium SKU.
```

docs/en/rules/Azure.ServiceBus.GeoReplica.md
```diff
@@ -0,0 +1,128 @@
...
+
+This feature is only supported for the `Premium` SKU.
+
+This feature is currently in preview.
+
+## LINKS
+
+- [RE:05 Redundancy](https://learn.microsoft.com/azure/well-architected/reliability/redundancy)
+- [Geo-replication](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-geo-replication)
+- [Configure geo-replication](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-geo-replication#setup)
```
BernieWhite, line 127:
```suggestion
- [Configure geo-replication](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-geo-replication#setup)
- [Migrate existing Azure Service Bus standard namespaces to the premium tier](https://learn.microsoft.com/azure/service-bus-messaging/service-bus-migrate-standard-premium)
```

BernieWhite (changes requested):
Thanks @BenjaminEngeset. Great work, only a few minor suggestions.

BenjaminEngeset:
Thank you for the feedback. The requested changes have been implemented, @BernieWhite.

BernieWhite (approved):

PR #2997, opened 2024-07-21
feat(new): Added Azure.AKS.MaintenanceWindow

### LINKED ISSUE #2444: Use planned maintenance feature with AKS clusters
### Existing rule

_No response_

### Suggested rule

A planned maintenance window should be used to schedule upgrades to avoid periods of high cluster utilization.

The `aksManagedAutoUpgradeSchedule` should be used for all cluster upgrade scenarios and `aksManagedNodeOSUpgradeSchedule` for all node OS security patching scenarios.

### Pillar

Reliability

### Additional context

- https://learn.microsoft.com/azure/aks/planned-maintenance

BenjaminEngeset:
## PR Summary

Fixes #2444

Added Azure.AKS.MaintenanceWindow.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


BenjaminEngeset:
Ready for review @BernieWhite. This one was not straight forward.

The reason for not wanting the `default` configuration is because they all can exist and by configuring default wrongly, you might get some unexpected behavior. The documentation recommends using these two explicitly, even tho all three can exist at the same time. Let me know what you think about this approach.

docs/en/rules/Azure.AKS.MaintenanceWindow.md
```diff
@@ -0,0 +1,234 @@
...
+      "durationHours": 4,
+      "utcOffset": "+00:00",
+      "startDate": "2024-07-15",
+      "startTime": "00:00"
+    }
+  },
+  "dependsOn": [
+    "[resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName'))]"
+  ]
+}
```
BernieWhite, line 136:
```suggestion
{
  "type": "Microsoft.ContainerService/managedClusters/maintenanceConfigurations",
  "apiVersion": "2024-03-02-preview",
  "name": "[format('{0}/{1}', parameters('clusterName'), 'aksManagedAutoUpgradeSchedule')]",
  "properties": {
    "maintenanceWindow": {
      "schedule": {
        "weekly": {
          "intervalWeeks": 1,
          "dayOfWeek": "Sunday"
        }
      },
      "durationHours": 4,
      "utcOffset": "+00:00",
      "startDate": "2024-07-15",
      "startTime": "00:00"
    }
  },
  "dependsOn": [
    "[resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName'))]"
  ]
}
```

docs/en/rules/Azure.AKS.MaintenanceWindow.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
```
BernieWhite, line 138:
```suggestion
For node OS auto-upgrades see the example:

```json
{
  "type": "Microsoft.ContainerService/managedClusters/maintenanceConfigurations",
  "apiVersion": "2024-03-02-preview",
  "name": "[format('{0}/{1}', parameters('clusterName'), 'aksManagedNodeOSUpgradeSchedule')]",
  "properties": {
    "maintenanceWindow": {
      "schedule": {
        "weekly": {
          "intervalWeeks": 1,
          "dayOfWeek": "Sunday"
        }
      },
      "durationHours": 4,
      "utcOffset": "+00:00",
      "startDate": "2024-07-15",
      "startTime": "00:00"
    }
  },
  "dependsOn": [
    "[resourceId('Microsoft.ContainerService/managedClusters', parameters('clusterName'))]"
  ]
}
```

```

File: docs/en/rules/Azure.AKS.MaintenanceWindow.md [hunk 3]
```diff
@@ -0,0 +1,234 @@
...
+### Configure with Bicep
+
+To deploy AKS clusters that pass this rule:
+
+- Deploy two maintenance configurations.
+ - Set the `name` to `aksManagedAutoUpgradeSchedule` and `aksManagedNodeOSUpgradeSchedule`.
+
+For example:
```
BernieWhite, line 146:
```suggestion
- Deploy maintenance configurations for cluster version and node OS auto-upgrades.
  - For cluster version auto-upgrades, set the `name` property to `aksManagedAutoUpgradeSchedule`.
  - For node OS auto-upgrades, set the `name` property to `aksManagedNodeOSUpgradeSchedule`.

For cluster version auto-upgrades see the example:
```

docs/en/rules/Azure.AKS.MaintenanceWindow.md
```diff
@@ -0,0 +1,234 @@
...
+      ssh: {
+        publicKeys: [
+          {
+            keyData: sshRSAPublicKey
+          }
+        ]
+      }
+    }
+  }
+}
```
BernieWhite, line 186:
```suggestion
```

docs/en/rules/Azure.AKS.MaintenanceWindow.md
```diff
@@ -0,0 +1,234 @@
...
+          dayOfWeek: 'Sunday'
+        }
+      }
+      durationHours: 4
+      utcOffset: '+00:00'
+      startDate: '2024-07-15'
+      startTime: '00:00'
+    }
+  }
+}
```
BernieWhite, line 224:
```suggestion
}
```

For node OS auto-upgrades see the example:

```bicep
resource aksManagedNodeOSUpgradeSchedule 'Microsoft.ContainerService/managedClusters/maintenanceConfigurations@2024-03-02-preview' = {
  parent: aks
  name: 'aksManagedNodeOSUpgradeSchedule'
  properties: {
    maintenanceWindow: {
      schedule: {
        weekly: {
          intervalWeeks: 1
          dayOfWeek: 'Sunday'
        }
      }
      durationHours: 4
      utcOffset: '+00:00'
      startDate: '2024-07-15'
      startTime: '00:00'
    }
  }
}
```

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
```diff
@@ -319,6 +319,44 @@ Rule 'Azure.AKS.AuditAdmin' -Ref 'AZR-000445' -Type 'Microsoft.ContainerService/
...
+    }
+
+    else {
+        if ($PSRule.TargetName -match 'aksManagedAutoUpgradeSchedule$|aksManagedNodeOSUpgradeSchedule$') {
+            return $Assert.Pass()
+        }
+
+        $Assert.Fail().Reason($LocalizedData.AKSMaintenanceWindow, $PSRule.TargetName)
+    }
+}
```
BernieWhite, line 358: Couldn't see specifically in the docs that using the `default` configuration is not recommended.
Only that `aksManagedAutoUpgradeSchedule` and `aksManagedNodeOSUpgradeSchedule` should always be set.

But if you see that let me know.

So I think we should not fail (ignore) if `default` is set.

```suggestion
Rule 'Azure.AKS.MaintenanceWindow' -Ref 'AZR-000446' -Type 'Microsoft.ContainerService/managedClusters' -Tag @{ release = 'GA'; ruleSet = '2024_09'; 'Azure.WAF/pillar' = 'Reliability'; } {
      $maintenanceConfigs = @(GetSubResources -ResourceType 'Microsoft.ContainerService/managedClusters/maintenanceConfigurations')
      
      $hasAutoUpgrade = $false
      $hasNodeUpgrade = $false

      foreach ($config in $maintenanceConfigs) {
          if ($config.name -match 'aksManagedAutoUpgradeSchedule$') {
              $hasAutoUpgrade = $true
          }
          elseif ($config.name -match 'aksManagedNodeOSUpgradeSchedule$') {
              $hasNodeUpgrade = $true
          }
      }

      if ($hasAutoUpgrade -and $hasNodeUpgrade) {
          return $Assert.Pass()
      }
      else {
          $Assert.Fail().Reason($LocalizedData.AKSMaintenanceWindow, $PSRule.TargetName)
      }
}
```

BernieWhite (changes requested):
Thanks @BenjaminEngeset. Great work! Yes, this one has a few complexities.

I like the approach; I don't think we should flag `default` though.

Please see my suggestions below.

src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1
BenjaminEngeset: > Couldn't see specifically in the docs that using the `default` configuration is not recommended. Only that `aksManagedAutoUpgradeSchedule` and `aksManagedNodeOSUpgradeSchedule` should always be set.
> 
> But if you see that let me know.
> 
> So I think we should not fail (ignore) if `default` is set.

You are right. I agree with you.

BenjaminEngeset:
Thank you for the excellent feedback and suggestions, @BernieWhite. I have now implemented the requested changes.

Given that our rule specifically targets scenarios where the child resource is nested within the parent resource, will this rule consistently fail for clusters where it’s declared outside the parent? This approach seems to be the only viable solution currently anyways.

BernieWhite:
> Thank you for the excellent feedback and suggestions, @BernieWhite. I have now implemented the requested changes.
> 
> Given that our rule specifically targets scenarios where the child resource is nested within the parent resource, will this rule consistently fail for clusters where it’s declared outside the parent? That will at least make it clear for customers explicitly, but might be confusing that the rule still fails after they declare it. This approach seems to be the only viable solution currently anyways.

PSRule automatically nests the sub-resource into the parent during expansion so it shouldn't be an issue, if you define a cluster then the expectation is somewhere in the same deployment (although could be a separate module) you also define maintenance config. If IaC will be flagged as an issue.

The case that's PSRule doesn't handle is if you define a cluster in code then separately through a different process you add maintenance config, like deploy if not exists case with Azure Policy. In this case IaC is going to flag an issue with the code, which could be suppressed or excluded.

BernieWhite (approved):
Thanks @BenjaminEngeset! All good to merge if you're happy with this explanation.

Also, I added the in-flight export to this PR.

BenjaminEngeset:
> > Thank you for the excellent feedback and suggestions, @BernieWhite. I have now implemented the requested changes.
> > Given that our rule specifically targets scenarios where the child resource is nested within the parent resource, will this rule consistently fail for clusters where it’s declared outside the parent? That will at least make it clear for customers explicitly, but might be confusing that the rule still fails after they declare it. This approach seems to be the only viable solution currently anyways.
> 
> PSRule automatically nests the sub-resource into the parent during expansion so it shouldn't be an issue, if you define a cluster then the expectation is somewhere in the same deployment (although could be a separate module) you also define maintenance config. If IaC will be flagged as an issue.
> 
> The case that's PSRule doesn't handle is if you define a cluster in code then separately through a different process you add maintenance config, like deploy if not exists case with Azure Policy. In this case IaC is going to flag an issue with the code, which could be suppressed or excluded.

Thanks for the information, @BernieWhite. Given this, I have a broader question about a design pattern we’ve been using for several years. Specifically, I’m curious about why, in certain cases, we provide the child as the value to `-TargetType`.

My understanding has been that PSRule wouldn’t correctly handle scenarios where the child is declared outside its parent, hence having to add the child as the value to `-TargetType` to cover these scenarios.

For reference, here’s an example:

https://github.com/Azure/PSRule.Rules.Azure/blob/df2d5ef1afbab648858744e1507450553bd068b1/src/PSRule.Rules.Azure/rules/Azure.AKS.Rule.ps1#L296-L306

Could you clarify this design?

BenjaminEngeset:
> Thanks @BenjaminEngeset! All good to merge if you're happy with this explanation.
> 
> Also, I added the in-flight export to this PR.

Thank you! All good.

BernieWhite:
> Azure.AKS.EphemeralOSDisk

Thanks for the question @BenjaminEngeset.

For the rule `Azure.AKS.MaintenanceWindow` we are checking the state of `Microsoft.ContainerService/managedClusters` by checking it's sub-resources.
- If the sub-resources exist in the designed configuration it passes.
- If the sub-resources don't exist then it fails because the Azure default don't configure these sub-resources.

For the rule `Azure.AKS.EphemeralOSDisk` we are checking the state of `Microsoft.ContainerService/managedClusters/agentPools` which probably (most of the time) is in the cluster deployment.
- If the agent pools are in the desired state they pass.
- If no agent pools exist, the rule still passes.

**So when will agentPools not be automatically nested under a parent resource by PSRule?**

You're `existing` resource is the main case :) an `existing` resource is not a deployment of a `Microsoft.ContainerService/managedClusters` resource. It is a reference to a cluster that was deployed at some point previously.

The deployment for this doesn't have to be in the same deployment where the agent pool is created. It could for all intents be manually deployed by someone clicking in the portal, as long as it exists at the time the agent pool deployment references it.

Because there is no cluster defined, PSRule won't nest the sub-resource, it will be left as is. That said, the reason this happens is not because you used an `existing` reference, it's because in the deployment the cluster is not defined at all. If you used `existing` in a child module but another modules part of the deployment defined the cluster it would be nested.

BenjaminEngeset:
> > Azure.AKS.EphemeralOSDisk
> 
> Thanks for the question @BenjaminEngeset.
> 
> For the rule `Azure.AKS.MaintenanceWindow` we are checking the state of `Microsoft.ContainerService/managedClusters` by checking it's sub-resources.
> 
> * If the sub-resources exist in the designed configuration it passes.
> * If the sub-resources don't exist then it fails because the Azure default don't configure these sub-resources, and as a result the cluster is not in the desired state.
> 
> For the rule `Azure.AKS.EphemeralOSDisk` we are checking the state of `Microsoft.ContainerService/managedClusters/agentPools` which probably (most of the time) is in the cluster deployment.
> 
> * If the agent pools are in the desired state they pass.
> * If no agent pools exist, the rule still passes.
> 
> **So when will agentPools not be automatically nested under a parent resource by PSRule?**
> 
> You're `existing` resource is the main case :) an `existing` resource is not a deployment of a `Microsoft.ContainerService/managedClusters` resource. It is a reference to a cluster that was deployed at some point previously.
> 
> The deployment for this doesn't have to be in the same deployment where the agent pool is created. It could for all intents be manually deployed by someone clicking in the portal, as long as it exists at the time the agent pool deployment references it.
> 
> Because there is no cluster defined, PSRule won't nest the sub-resource, it will be left as is. That said, the reason this happens is not because you used an `existing` reference, it's because in the deployment the cluster is not defined at all. If you used `existing` in a child module but another modules part of the deployment defined the cluster it would be nested.

Thank you for taking your time to give this explanation, @BernieWhite. Truly grateful for that. It makes sense.

PR #3004, opened 2024-08-02
feat(new): Added Azure.VNET.NAT

### LINKED ISSUE #3003: [RULE] Use Azure NAT gateway for outbound access for virtual machines
### Existing rule

_No response_

### Suggested rule

By default, virtual machines (VMs) created in a virtual network without explicit outbound connectivity are assigned a default outbound public IP address. This IP address enables outbound connectivity to the internet.

Default Routing Behavior:
The subnet has a system default route that routes traffic with the destination 0.0.0.0/0 to the internet.
Normally, when this route is chosen, the traffic is NATed using either a default public IP address or the VM’s public IP address (if it exists) before being sent to the internet.

Behavior with NAT Gateway:
Once a NAT gateway is associated with the subnet, the behavior changes. Although the system default route will still have the next hop set to the internet, traffic is first sent to the NAT gateway.
The NAT gateway receives the traffic from the VMs in the subnet, NATs the traffic using its own public IP address, and then sends it to the internet.

The recommended method is to use NAT gateway as the explicit method of public connectivity for virtual machines to ensure security, explicit connectivity, and stability.

Also highly relevant for AKS clusters.

### Pillar

Security

### Additional context

- [SE:06 Network controls](https://learn.microsoft.com/azure/well-architected/security/networking)
- [Azure NAT Gateway](https://learn.microsoft.com/azure/nat-gateway/nat-overview)
- [Associate a NAT gateway to the subnet](https://learn.microsoft.com/azure/load-balancer/load-balancer-outbound-connections#2-associate-a-nat-gateway-to-the-subnet)
- [Default outbound access](https://learn.microsoft.com/azure/virtual-network/ip-services/default-outbound-access)
- [Azure deployment reference - Virtual Network](https://learn.microsoft.com/azure/templates/microsoft.network/virtualnetworks)
- [Azure deployment reference - Subnet](https://learn.microsoft.com/azure/templates/microsoft.network/virtualnetworks/subnets)

BenjaminEngeset:
## PR Summary

Fixes #3003

Added Azure.VNET.NAT.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -33,6 +33,9 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
   - Azure Kubernetes Service:
     - Verify that clusters have the customer-controlled maintenance windows 'aksManagedAutoUpgradeSchedule' and 'aksManagedNodeOSUpgradeSchedule' configured by @BenjaminEngeset.
       [#2444](https://github.com/Azure/PSRule.Rules.Azure/issues/2444)
+  - Virtual Network:
+    - Verify that subnets have NAT gateway associated by @BenjaminEngeset.
+      [#3003](https://github.com/Azure/PSRule.Rules.Azure/issues/3003)
 
 What's changed since pre-release v1.39.0-B0009:
 
--- a/docs/en/rules/Azure.VNET.NAT.md
+++ b/docs/en/rules/Azure.VNET.NAT.md
@@ -1,0 +1,157 @@
+---
+severity: Critical
+pillar: Security
+category: SE:06 Network controls
+resource: Virtual Network
+online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.VNET.NAT/
+---
+
+# Outbound access
+
+## SYNOPSIS
+
+Use Azure NAT gateway for outbound access for virtual machines.
+
+## DESCRIPTION
+
+By default, virtual machines (VMs) created in a virtual network without explicit outbound connectivity are assigned a default outbound public IP address. This IP address enables outbound connectivity to the internet.
+
+Default Routing Behavior:
+The subnet has a system default route that routes traffic with the destination 0.0.0.0/0 to the internet.
+Normally, when this route is chosen, the traffic is NATed using either a default public IP address or the VM’s public IP address (if it exists) before being sent to the internet.
+
+Behavior with NAT Gateway:
+Once a NAT gateway is associated with the subnet, the behavior changes. Although the system default route will still have the next hop set to the internet, traffic is first sent to the NAT gateway.
+The NAT gateway receives the traffic from the VMs in the subnet, NATs the traffic using its own public IP address, and then sends it to the internet.
+
+Why use a Azure NAT gateway?
+
+- Enhanced Security
+  - Zero Trust Principles: The NAT gateway adheres to the zero trust security model, reducing the attack surface by not exposing your VMs directly to the internet. It ensures that outbound connections are initiated securely without inbound ports being open.
+	- Managed Outbound Connections: By centralizing outbound connections through a NAT gateway, you reduce the complexity of managing individual public IPs and security rules on each VM.
+- Consistent and Predictable IP Management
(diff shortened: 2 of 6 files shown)


BenjaminEngeset:

The rule is initially created for used-managed virtual machines, but there are other services that can take advantage of this as well.

BernieWhite:
@BenjaminEngeset this one is a little more nuanced.

When a customer starts deploying multiple workloads for most cases they should be thinking about using a VWAN or hub and spoke network with a central Azure Firewall or NVA for controlling east/ west and north/ south traffic, following something like the cloud adoption framework. As a result, every subnet shouldn't have a NAT gateway attached. Realistically the number of internet egress points should be minimized.

You could say that anything in the VNET hub should have a NAT gateway, but even then, there is cases when this is not ideal.

- NAT Gateway is zonal, so you should deploy one in each AZ you deploy to, but with something like Azure Firewall that supports zone redundancy deploying a single NAT gateway (only one can be deployed in a subnet) could decrease the availability of outbound traffic if the zone that failed was the same as the NAT gateway.
- In the NVA configuration, there is often a private and public subnet used in-front and behind the NVA instances. The NAT gateway should only be deployed on the public subnet.

So, I'm not sure if we can reliability detect the cases from code the cases that should use a NAT gateway.

Did you have any further thoughts on this?

BenjaminEngeset:
> @BenjaminEngeset this one is a little more nuanced.
> 
> When a customer starts deploying multiple workloads for most cases they should be thinking about using a VWAN or hub and spoke network with a central Azure Firewall or NVA for controlling east/ west and north/ south traffic, following something like the cloud adoption framework. As a result, every subnet shouldn't have a NAT gateway attached. Realistically the number of internet egress points should be minimized.
> 
> You could say that anything in the VNET hub should have a NAT gateway, but even then, there is cases when this is not ideal.
> 
> * NAT Gateway is zonal, so you should deploy one in each AZ you deploy to, but with something like Azure Firewall that supports zone redundancy deploying a single NAT gateway (only one can be deployed in a subnet) could decrease the availability of outbound traffic if the zone that failed was the same as the NAT gateway.
> * In the NVA configuration, there is often a private and public subnet used in-front and behind the NVA instances. The NAT gateway should only be deployed on the public subnet.
> 
> So, I'm not sure if we can reliability detect the cases from code the cases that should use a NAT gateway.
> 
> Did you have any further thoughts on this?

I agree with you @BernieWhite and I am truly grateful for your honest feedback.

What I can do instead is to create a new issue and new rule regarding `AzureFirewallSubnet` that is towards mitigating port exhaustion, so zonal-deployed firewalls can use a NAT gateway, if needing more than `1248000` connections at a time for an instance.

What do you think about that? There is official documentation for this.

https://learn.microsoft.com/en-us/azure/firewall/integrate-with-nat-gateway

BernieWhite:
> > @BenjaminEngeset this one is a little more nuanced.
> > When a customer starts deploying multiple workloads for most cases they should be thinking about using a VWAN or hub and spoke network with a central Azure Firewall or NVA for controlling east/ west and north/ south traffic, following something like the cloud adoption framework. As a result, every subnet shouldn't have a NAT gateway attached. Realistically the number of internet egress points should be minimized.
> > You could say that anything in the VNET hub should have a NAT gateway, but even then, there is cases when this is not ideal.
> > 
> > * NAT Gateway is zonal, so you should deploy one in each AZ you deploy to, but with something like Azure Firewall that supports zone redundancy deploying a single NAT gateway (only one can be deployed in a subnet) could decrease the availability of outbound traffic if the zone that failed was the same as the NAT gateway.
> > * In the NVA configuration, there is often a private and public subnet used in-front and behind the NVA instances. The NAT gateway should only be deployed on the public subnet.
> > 
> > So, I'm not sure if we can reliability detect the cases from code the cases that should use a NAT gateway.
> > Did you have any further thoughts on this?
> 
> I agree with you @BernieWhite and I am truly grateful for your honest feedback.
> 
> What I can do instead is to create a new issue and new rule regarding `AzureFirewallSubnet` that is towards mitigating port exhaustion and less administrative overhead, so zonal-deployed firewalls can use a NAT gateway What do you think about that? There is official documentation for this.
> 
> https://learn.microsoft.com/en-us/azure/firewall/integrate-with-nat-gateway

Yes #3005 #3006 is a good idea.

PR #3006, opened 2024-08-04
feat(new): Added Azure.VNET.FirewallSubnetNAT

### LINKED ISSUE #3005: [RULE]  Zonal-deployed Azure Firewalls should consider using an Azure NAT Gateway for outbound access
### Existing rule

_No response_

### Suggested rule

Azure Firewall can be deployed with up to 250 public IP addresses, each providing 2,496 SNAT ports. This setup offers a maximum of 1,248,000 SNAT ports.

Managing a large number of public IP addresses comes with challenges, particularly regarding downstream IP address filtering requirements. When Azure Firewall is associated with multiple public IP addresses, these filtering requirements must be applied to all associated addresses. 
Even when using Public IP address prefixes, associating 250 public IP addresses requires managing 16 public IP address prefixes on the downstream side.

A more efficient solution for scaling and dynamically allocating outbound SNAT ports is to use an Azure NAT Gateway:

- High Capacity: Each public IP address on a NAT Gateway provides 64,512 SNAT ports, and up to 16 public IP addresses can be associated, resulting in up to 1,032,192 SNAT ports.
- Dynamic Allocation: SNAT ports are dynamically allocated at the subnet level, making all provided SNAT ports available on demand for outbound connectivity.

This configuration simplifies management for downstream systems, as it requires handling only up to 16 public IP addresses.

### Pillar

Reliability

### Additional context

- [RE:05 Redundancy](https://learn.microsoft.com/azure/well-architected/reliability/redundancy)
- [Scale SNAT ports with Azure NAT Gateway](https://learn.microsoft.com/azure/firewall/integrate-with-nat-gateway)
- [Plan for inbound and outbound internet connectivity](https://learn.microsoft.com/azure/cloud-adoption-framework/ready/azure-best-practices/plan-for-inbound-and-outbound-internet-connectivity)
- [Azure deployment reference - Virtual Network](https://learn.microsoft.com/azure/templates/microsoft.network/virtualnetworks)
- [Azure deployment reference - Subnet](https://learn.microsoft.com/azure/templates/microsoft.network/virtualnetworks/subnets)

BenjaminEngeset:
## PR Summary

Fixes #3005

Added Azure.VNET.FirewallSubnetNAT.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -117,6 +117,7 @@
     "SBOM",
     "SIEM",
     "SKUs",
+    "SNAT",
     "stateful",
     "subnet",
     "subnets",
--- a/docs/CHANGELOG-v1.md
+++ b/docs/CHANGELOG-v1.md
@@ -33,6 +33,9 @@ See [upgrade notes][1] for helpful information when upgrading from previous vers
   - Azure Kubernetes Service:
     - Verify that clusters have the customer-controlled maintenance windows 'aksManagedAutoUpgradeSchedule' and 'aksManagedNodeOSUpgradeSchedule' configured by @BenjaminEngeset.
       [#2444](https://github.com/Azure/PSRule.Rules.Azure/issues/2444)
+  - Virtual Network:
+    - Verify that zonal-deployed Azure firewalls uses Azure NAT Gateway for outbound access by @BenjaminEngeset.
+      [##3005](https://github.com/Azure/PSRule.Rules.Azure/issues/#3005)
 
 What's changed since pre-release v1.39.0-B0009:
 
--- a/docs/en/rules/Azure.VNET.FirewallSubnetNAT.md
+++ b/docs/en/rules/Azure.VNET.FirewallSubnetNAT.md
@@ -6,7 +6,7 @@ resource: Virtual Network
 online version: https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.VNET.FirewallSubnetNAT/
 ---
 
-# Outbound access
+# Use NAT gateway with Azure Firewall for outbound access
 
 ## SYNOPSIS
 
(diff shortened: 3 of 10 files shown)


BenjaminEngeset:
Ready for review, @BernieWhite. Let me know what you think. If you do not see this as relevant to merge into main, that’s fine.

BernieWhite (commented):
Hi @BenjaminEngeset. Looks good. We have to be careful about generating a false positive here, because for most customers are probably deploying a zone redundant configuration without a NAT gateway and that is still a good configuration.

Requiring a NAT gateway is likely to be the exception case. As a result, I think using a configuration value `AZURE_FIREWALL_IS_ZONAL` is an ideal path, since we can't detect if the Azure Firewall is deployed zonal from code.

I've added some updates to this end. What do you think? Are you happy with the updates?

BenjaminEngeset:
> Hi @BenjaminEngeset. Looks good. We have to be careful about generating a false positive here, because for most customers are probably deploying a zone redundant configuration without a NAT gateway and that is still a good configuration.
> 
> Requiring a NAT gateway is likely to be the exception case. As a result, I think using a configuration value `AZURE_FIREWALL_IS_ZONAL` is an ideal path, since we can't detect if the Azure Firewall is deployed zonal from code.
> 
> I've added some updates to this end. What do you think? Are you happy with the updates?

I agree with you and thanks for the great updates @BernieWhite. I am very pleased with this!

BernieWhite:
> > Hi @BenjaminEngeset. Looks good. We have to be careful about generating a false positive here, because for most customers are probably deploying a zone redundant configuration without a NAT gateway and that is still a good configuration.
> > Requiring a NAT gateway is likely to be the exception case. As a result, I think using a configuration value `AZURE_FIREWALL_IS_ZONAL` is an ideal path, since we can't detect if the Azure Firewall is deployed zonal from code.
> > I've added some updates to this end. What do you think? Are you happy with the updates?
> 
> I agree with you and thanks for the great updates @BernieWhite. I am very pleased with this!

There is a bug in my code let me fix that.

BernieWhite (approved):

PR #3022, opened 2024-08-15
Added Azure.VM.ASDistributeTraffic

### LINKED ISSUE #67: VMs in an availablity set should use a load balancer or application gateway
# Rule request

## Suggested rule change

VMs in an availability set should use a load balancer or application gateway to distribute traffic as part of a backend pool.

## Applies to the following

- Resource type: **Microsoft.Compute/virtualMachines**

BenjaminEngeset:
## PR Summary

Fixes #67

Added Azure.VM.ASDistributeTraffic.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [ ] Link to a filed issue


BenjaminEngeset:
Ready for review, @BernieWhite.

This one is a bit special, as in all honesty, configuring virtual network card interfaces within the resource itself is not so common and should not be recommended really. Cannot even decide the full name for the NIC. It's not possible to configure `properties.networkProfile.networkInterfaceConfigurations` and `properties.networkProfile.networkInterfaces` at the same time within a template either.

After the deployment by using `properties.networkProfile.networkInterfaceConfigurations`,  the property is gone in the template and the `properties.networkProfile.networkInterfaces` is created with the NICs that the interface configuration created.

So the rule is not really that useful for scenarios when exporting.

Let me know what you think. This is not an ideal rule, but I think it's not possible to do anything else?

BernieWhite:
> 'Microsoft.Network/networkInterfaces'

Only for in-flight cases.

BenjaminEngeset:
> > 'Microsoft.Network/networkInterfaces'
> 
> Only for in-flight cases. But this rule is only going to be applicable for in-flight cases.

Gotcha. Update from me soon, I will notify you when I am done.

BenjaminEngeset:

As it is now each IP configuration requires to reference a pool, otherwise it will fail.

The rule can also be designed so that one IP configuration referencing a pool is enough.

PR #3818, opened 2026-05-26
add support for roledefinitions function

### LINKED ISSUE #3776: [FEATURE] Add support for roleDefinitions functions
### Your suggestion

The `roleDefinitions` function has been [added to Bicep](https://github.com/Azure/bicep/releases/tag/v0.42.1), and would be good to have support by PSRule.

### Alternatives

na

### Additional context

Original issue https://github.com/microsoft/PSRule/issues/3325

polatengin:
## PR Summary

- Add support for the Bicep-emitted `roleDefinitions(...)` ARM expression function.
- Return the expected `id` and `roleDefinitionId` fields so expressions like `roleDefinitions('Contributor').id` can expand correctly.
- Add focused unit coverage for direct function evaluation and expression-builder property access.
- Update the changelog.

Fixes #3776

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
- **Other code changes**
  - [x] Link to a filed issue

### DIFF
--- a/src/PSRule.Rules.Azure/Arm/Expressions/Functions.cs
+++ b/src/PSRule.Rules.Azure/Arm/Expressions/Functions.cs
@@ -48,6 +48,7 @@ internal static class Functions
 
     private const string FUNCTION_LIST_WITH_SECURE_OUTPUTS = "listOutputsWithSecureValues";
     private const string RESOURCE_TYPE_ROLE_DEFINITIONS = "Microsoft.Authorization/roleDefinitions";
+    private const string ROLE_DEFINITION_ID_PLACEHOLDER = "00000000-0000-0000-0000-000000000000";
 
     private const string FORMAT_ISO8601 = "yyyy-MM-ddTHH:mm:ssZ";
     private const string FORMAT_AZURE_DATETIME = "M/d/yyyy h:mm:ss tt";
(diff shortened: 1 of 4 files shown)


tests/PSRule.Rules.Azure.Tests/Arm/Expressions/ExpressionBuilderTests.cs
```diff
@@ -149,6 +149,16 @@ public void JsonQuoteNesting()
...
     }
 
+    [Fact]
+    public void BuildExpressionWithRoleDefinitions()
+    {
+        var context = GetContext();
+
+        var actual = Build(context, "[roleDefinitions('Contributor').id]") as string;
+
+        Assert.Equal("/subscriptions/ffffffff-ffff-ffff-ffff-ffffffffffff/providers/Microsoft.Authorization/roleDefinitions/Contributor", actual);
```
BernieWhite: Result is JValue.

```suggestion
        var actual = Build(context, "[roleDefinitions('Contributor').id]") as JValue;

        Assert.Equal("/subscriptions/ffffffff-ffff-ffff-ffff-ffffffffffff/providers/Microsoft.Authorization/roleDefinitions/Contributor", actual.Value<string>());
```

src/PSRule.Rules.Azure/Arm/Expressions/Functions.cs
```diff
@@ -1272,6 +1277,25 @@ internal static object ManagementGroupResourceId(ITemplateContext context, objec
         return ResourceHelper.CombineResourceId(managementGroupName, resourceType, name);
     }
 
+    /// <summary>
+    /// roleDefinitions(roleName)
+    /// </summary>
+    internal static object RoleDefinitions(ITemplateContext context, object[] args)
+    {
```
BernieWhite, line 1285: Some thoughts:
- The Bicep implementation, would the resource ID be scoped to a management group if the deployment scope was a management group? If so, returning a subscription scoped resource ID might be an issue,
- Based on the documentation `id` and `roleDefinitionId` would both contain the GUID of the built-in role definition. So we should ideally return a GUID instead of the role name. Two options:
  - We could used use a placeholder like `00000000-0000-0000-0000-000000000000`.
  - We could ingest a list of known built-in roles and their mappings from `data/`, then fall back to a placeholder for anything unknown at the time of the call. This means we'd need to export the data and build it into the binary, however this would be the most compatible. Why? It's hard to say how a customer would use the results in code. `id` fairly easy, but `roleDefinitionId` as a lookup into a map might break.

BernieWhite (changes requested):
Thanks @polatengin. I've added a few comments below.

src/PSRule.Rules.Azure/Arm/Expressions/Functions.cs
polatengin: it's a great point, I've updated the implementation 👍

BernieWhite (approved):
Looks great @polatengin. Thanks for the updates.

PR #3863, opened 2026-07-14
feat(fleet): add Azure.Fleet.SecureBoot rule (AZR-000545)

### LINKED ISSUE #3729: [RULE] Azure Fleets should use secure boot
### Existing rule

_No response_

### Suggested rule

Fleets `Microsoft.AzureFleet/fleets` should use secure boot by setting the property `properties.computeProfile.baseVirtualMachineProfile.securityProfile.securityType` = `TrustedLaunch` and `properties.computeProfile.baseVirtualMachineProfile.securityProfile.uefiSettings.secureBootEnabled` = `true`.

At the root of Trusted Launch is Secure Boot for your VM. Secure Boot, which is implemented in platform firmware, protects against the installation of malware-based rootkits and boot kits. Secure Boot works to ensure that only signed operating systems and drivers can boot. It establishes a "root of trust" for the software stack on your VM.

With Secure Boot enabled, all OS boot components (boot loader, kernel, kernel drivers) require trusted publishers signing. Both Windows and select Linux distributions support Secure Boot. If Secure Boot fails to authenticate that the image is signed with a trusted publisher, the VM fails to boot.

### Pillar

Security

### Additional context

Setting the security type to `ConfidentialVM` is also acceptable.

- https://learn.microsoft.com/azure/virtual-machines/trusted-launch
- https://learn.microsoft.com/azure/templates/microsoft.azurefleet/fleets#securityprofile

alexchen-sys:
## PR Summary

Fixes: #3729

Adds `Azure.Fleet.SecureBoot` so Azure Fleet base VM profiles require Trusted Launch or Confidential VM with Secure Boot enabled.

Path checked:
- `properties.computeProfile.baseVirtualMachineProfile.securityProfile.securityType` ∈ `TrustedLaunch` | `ConfidentialVM`
- `properties.computeProfile.baseVirtualMachineProfile.securityProfile.uefiSettings.secureBootEnabled` = `true`

Mirrors existing `Azure.VM.SecureBoot` / `Azure.VMSS.SecureBoot` guidance for fleet profiles.

## PR Checklist

- [x] Summarized changes
- [x] Change is not breaking
- **Rule changes**
  - [x] Link to a filed issue
  - [x] [Change log](https://github.com/Azure/PSRule.Rules.Azure/blob/main/docs/changelog.md) has been updated with change under unreleased section

### DIFF
--- a/data/resource-type-mapping.json
+++ b/data/resource-type-mapping.json
@@ -352,20 +352,20 @@
     "displayName": "Enable AKS Container insights"
   },
   {
-    "severity": "Awareness",
-    "pillar": "Operational Excellence",
-    "category": "OE:04 Continuous integration",
-    "resourceType": "Microsoft.ContainerService/managedClusters",
-    "ruleId": "Azure.AKS.DNSPrefix",
-    "displayName": "Use valid AKS cluster DNS prefix"
-  },
-  {
     "severity": "Important",
     "pillar": "Security",
     "category": "Azure resources",
     "resourceType": "Microsoft.ContainerService/managedClusters",
     "ruleId": "Azure.AKS.DefenderProfile",
     "displayName": "Enable Defender profile"
+  },
+  {
+    "severity": "Awareness",
+    "pillar": "Operational Excellence",
+    "category": "OE:04 Continuous integration",
+    "resourceType": "Microsoft.ContainerService/managedClusters",
+    "ruleId": "Azure.AKS.DNSPrefix",
+    "displayName": "Use valid AKS cluster DNS prefix"
   },
   {
     "severity": "Important",
@@ -632,6 +632,22 @@
     "displayName": "Kubernetes Cluster version is old"
   },
   {
+    "severity": "Important",
+    "pillar": "Cost Optimization",
+    "category": "CO:06 Usage and billing increments",
+    "resourceType": "Microsoft.Insights/scheduledQueryRules",
+    "ruleId": "Azure.Alert.HighFrequencyQuery",
+    "displayName": "Scheduled Query Alert is configured to use a high frequency query"
+  },
+  {
+    "severity": "Important",
+    "pillar": "Cost Optimization",
+    "category": "CO:13 Personnel time",
+    "resourceType": "Microsoft.Insights/metricAlerts",
+    "ruleId": "Azure.Alert.MetricAutoMitigate",
+    "displayName": "Metric Alert requires manual mitigation"
+  },
+  {
     "severity": "Awareness",
     "pillar": "Operational Excellence",
     "category": "OE:04 Tools and processes",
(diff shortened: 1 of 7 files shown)


tests/PSRule.Rules.Azure.Tests/Azure.Fleet.Tests.ps1
```diff
@@ -53,8 +53,24 @@ Describe 'Azure.Fleet' -Tag 'Fleet' {
...
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult.Length | Should -Be 3;
+            $ruleResult.TargetName | Should -BeIn 'fleet-001', 'fleet-002', 'fleet-003';
+
+            # Pass
+            $ruleResult = @($filteredResult | Where-Object { $_.Outcome -eq 'Pass' });
+            $ruleResult | Should -Not -BeNullOrEmpty;
+            $ruleResult.Length | Should -Be 2;
+            $ruleResult.TargetName | Should -BeIn 'fleet-004', 'fleet-005';
         }
```
polatengin, line 74: Could you add separate test cases for the two checks? The three failing fleets are missing the whole security profile, so they fail both checks at the same time. The test still passes if the `secureBootEnabled` assertion is removed.

Please add these cases:

- `TrustedLaunch` with `secureBootEnabled: false` should fail.
- An invalid or missing `securityType` with `secureBootEnabled: true` should fail.
- `ConfidentialVM` with `secureBootEnabled: true` should pass.

This will confirm that both properties are checked and that `ConfidentialVM` is supported.

data/resource-type-mapping.json
polatengin, line 1: Is the full update to this generated mapping file needed in this PR? It changes 920 lines, but when the entries are compared without considering their order, the only new entry is `Azure.Fleet.SecureBoot`. What I understand from recent rule PRs have left this file to the separate mapping maintenance process.

@BernieWhite should we remove this file from the PR, or explain why it needs to be included?

What you think?

polatengin (changes requested):
The rule matches issue #3729 and follows the existing VM and VMSS Secure Boot rules. The rule ID, `2026_09` rule set, metadata, messages, documentation, examples, and changelog all look correct.

I have one request about test coverage and one question about the generated mapping file.

alexchen-sys:
Thanks for the review — agreed on both points.

Updates on the latest head:
1. **Tests** — added separate fleets so the two checks are not only covered together:
   - `fleet-006`: `TrustedLaunch` + `secureBootEnabled: false` → Fail
   - `fleet-007`: missing `securityType` + `secureBootEnabled: true` → Fail
   - `fleet-008`: `ConfidentialVM` + `secureBootEnabled: true` → Pass
2. **`data/resource-type-mapping.json`** — removed the generated churn from this PR and restored `main`. Left mapping to the normal maintenance process.

Happy to adjust further if you’d like different fixture names/layout.

polatengin:
> Thanks for the review — agreed on both points.

Can you also rebase your PR branch with the main, please?

alexchen-sys:
Rebased onto latest `main`.

Head is now `28db76665108` (clean rebase, no other changes).

